PR #6, opened 2022-04-01
Modularize app

nesium:
In this branch I've modularized the app in a way that has been explained by [pointfree here](https://www.pointfree.co/episodes/ep171-modularization-part-1).

This will keep entanglement at bay as the code base grows, helps with testability, allows to build very narrow preview apps for iterating on features and reduces compile times.

Unfortunately modularization often breaks Xcode Previews. Right now they still work but cannot access assets when run from a module. IMO this is a fair price to pay because in my experience Xcode Previews are fickle and break sooner or later as the project grows, at least in their current state. Focused preview apps are a good replacement for that.

The assets themselves should be moved into the Assets module at a later point so that preview apps can access them. And the code for accessing colors and localized strings could preferably be generated with [swiftgen](https://github.com/SwiftGen/SwiftGen).


valeriansaliou:
Hey Marc,

Thanks a tons for that! I've stacked the article you've sent about Modularization, for a later read.

On Xcode Previews, I do agree w/ you they are fickle. It's also super easy to mistakenly create infinite loops or bogous allocations resulting in a memleak in your code while live-previewing, and Xcode would happily create a preview process that'd hog on the CPU or RAM, until the OS kills it because it's taking all resources. That was my main issue w/ it while coding this static version of Prose.

nesium:
This popped up today and might be an interesting alternative to Xcode Previews: https://merowing.info/2022/04/hot-reloading-in-swift/

valeriansaliou:
If this doesnt add too much runtime costs on the final app, I think we should go for it yes. I share your pain on the development cycle w/o such things.

RemiBardon:
Seeing the project structure, I just realized I know how to fix your preview problem 💡

I think what you're experiencing is https://github.com/RemiBardon/xcode-preview-spm-resources-bug, and you can fix it by using [this workaround](https://github.com/RemiBardon/xcode-preview-spm-resources-bug/commit/6599c31af1c49fe884e2b84149043aee51aca031).

@nesium What do you think? Does it fix the problem here?

nesium:
Thanks, @RemiBardon! I'll give it a try.

PR #7, opened 2022-04-04
Authentication feature

nesium:
This PR adds a very basic login screen that uses the [composable architecture](https://github.com/pointfreeco/swift-composable-architecture) (TCA) for state management.

For verifying the credentials (this is just a static check ATM) the app calls into a Rust library via [UniFFI](https://mozilla.github.io/uniffi-rs/), which was the primary goal. I couldn't get the prose-core-client to compile, so I created a [small project for testing](https://github.com/nesium/uniffi_test). I've added a script which builds the Rust library and code generated by UniFFI into a XCFramework. Currently the script only builds the slice for M1 Macs and is most likely not the best way to do it, but one problem at a time.

As mentioned the main intent for this PR is to get the ball rolling so that we can figure out how we want to integrate the Rust library within the app and communicate between the two.

To try:
1. Install uniffi-bindgen (`$ cargo install uniffi_bindgen`)
2. Checkout [uniffi_test](https://github.com/nesium/uniffi_test) next to the prose-app-macos folder.
3. Build the Rust library by changing into the uniffi_test directory and running `$ make`.
4. Build the XCFramework by changing into the prose-app-macos folder and running `$ make`.
5. Build and run the app in Xcode.
6. You'll see a login form with prefilled data. You can login by clicking the Login button with that data and will see the jid returned from Rust in the sidebar footer. If you set the Jid to anything other than "hello@prose.org", you'll see the error thrown from Rust "The user does not exist" after tapping the Login button. If you change the password to anything other than "password" you'll see an error "Invalid credentials".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,6 @@ fastlane/test_output
 # https://github.com/johnno1962/injectionforxcode
 
 iOSInjectionProject/
+
+Build/
+ProseCore/ProseCore/ProseCore.xcframework
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,37 @@
+RUST_LIB_PATH = ../uniffi_test
+RUST_LIB_NAME = uniffi_test
+SWIFT_LIB_NAME = ProseCore
+BUILD_FOLDER = Build
+
+all: interface framework
+
+interface:
+	uniffi-bindgen generate $(RUST_LIB_PATH)/src/$(RUST_LIB_NAME).udl -o ./$(BUILD_FOLDER)/interface --language swift
+	sed -i '' 's/import uniffi_testFFI/@_implementationOnly import uniffi_testFFI/g' ./$(BUILD_FOLDER)/interface/$(RUST_LIB_NAME).swift
+
+framework:
+	rm -f $(BUILD_FOLDER)/lib$(SWIFT_LIB_NAME).a
+	rm -f $(BUILD_FOLDER)/$(SWIFT_LIB_NAME).a
+	rm -f $(BUILD_FOLDER)/$(SWIFT_LIB_NAME).swiftmodule
+	
+	(cd $(BUILD_FOLDER); swiftc \
+		-module-name $(SWIFT_LIB_NAME) \
+		-emit-library -o lib$(SWIFT_LIB_NAME).a \
+		-emit-module -emit-module-path . \
+		-parse-as-library \
+		-L $(RUST_LIB_PATH)/target/aarch64-apple-darwin/release \
+		-l$(RUST_LIB_NAME) \
+		-Xcc -fmodule-map-file=./interface/$(RUST_LIB_NAME)FFI.modulemap \
+		-static \
+		-enable-library-evolution \
+		./interface/$(RUST_LIB_NAME).swift)
+	
+	libtool -static -o $(BUILD_FOLDER)/$(SWIFT_LIB_NAME).a $(BUILD_FOLDER)/lib$(SWIFT_LIB_NAME).a $(RUST_LIB_PATH)/target/aarch64-apple-darwin/release/lib$(RUST_LIB_NAME).a
+	
+	rm -rf ProseCore/ProseCore/$(SWIFT_LIB_NAME).xcframework
+	
+	xcodebuild -create-xcframework \
+		-library ./Build/$(SWIFT_LIB_NAME).a \
+		-output ProseCore/ProseCore/$(SWIFT_LIB_NAME).xcframework
+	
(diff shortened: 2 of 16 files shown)


valeriansaliou:
Thanks, Marc! LGTM.

Curious though, which issues did you encounter w/ the Rust library? Was it when building it independantly and running an example (it's very minimal for now)? Or rather, when trying to bind it to the macOS app? (due to it being compiled w/ Rust std libs maybe?)

nesium:
When I try to build prose-core-client I'm seeing the same problems as [CI does](https://github.com/prose-im/prose-core-client/runs/5606619653?check_suite_focus=true).

This is something in the xmpp-parsers crate. It works with the unstable toolchain, however with unstable xmpp-rs doesn't build.

I think the problem is that these are outdated creates as both libraries have [been merged](https://gitlab.com/xmpp-rs/tokio-xmpp) into [this repo](https://gitlab.com/xmpp-rs/xmpp-rs), which seems to be different from the [xmpp-rs crate](https://crates.io/crates/xmpp-rs) that has last been updated 5 years ago.

valeriansaliou:
Hey Marc,

Yes, I need to spend a bit more time on the library, it's just not ready yet, but you may want to look into it as well when we start working on this in more details.

The state of XMPP in Rust is not really there yet, although xmpp-rs on https://gitlab.com/xmpp-rs/xmpp-rs is already pretty usable, w/ https://github.com/paulfariello/aparte using it.

RemiBardon:
Hello Marc 👋🏻

Are we going to use The Composable Architecture for the Prose applications (macOS / iOS)? I am quite experienced with it, as I've been using the library for many months now (on iOS). I can't wait to help you 🙂

(In case you didn't know, I will be working with you for 3 months this Summer 👨🏻‍💻)

nesium:
Hi Remi. Good to hear from you! Valerian of course mentioned that we'll be working together. 

That's great that you're already experienced with The Composable Architecture! Yes, I was looking to use the TCA in Prose as it worked really well for me in the past, for testability and modularizing an app. I hope you enjoyed working with the library as well?

Looking forward to summer then! 🌞

RemiBardon:
Yes, I really enjoy it, as it is very powerful 🙂

The main problems I have are animations (navigation and lists) and integration with Core Data.

I havent spent much time trying to work around the first, as I'm not currently working on it. From what I've seen from Point-Free, I think the problem is mainly the lack of API in SwiftUI (not sure).

For Core Data, I'm kind of duplicating a lot of data structures… but it works… and I think it's how it's supposed to be 🤔

nesium:
I'm glad to hear that you enjoy it! I haven't spent too much time with SwiftUI yet to be honest, so I don't know anything about that problem really. Combining TCA and DiffableDataSources for animating changes in lists worked well for me though.

You mean duplication of data structures probably because TCA state requires value types whereas NSManagedObjects are reference types? Personally, I've dropped CoreData a good while ago for [GRDB](https://github.com/groue/GRDB.swift) which works well with value types and never looked back 😬.

RemiBardon:
> I haven't spent too much time with SwiftUI yet to be honest
Oh so you've use TCA with UIKit? I haven't tried, I have no idea how it integrates 🤔 I used UIKit for some time, and switched to SwiftUI when it was released, so TCA didn't exist at the time.

> TCA state requires value types whereas NSManagedObjects are reference types?
Yes

> Personally, I've dropped CoreData a good while ago for GRDB which works well with value types and never looked back 😬.
I didn't know about GRDB, I'll have a look someday. I have two questions in mind:
- Core Data has a performance-oriented feature called "faults". It's basically the fact that every object member can be `nil`, and lazily fetched when you need it. This allows `SELECT`ing less columns in the SQL queries. I wonder how you handle this with value types. Do you just don't care as it's probably an optimization you don't need at first or do you create as many data structures as you need for your views? (Or you use the structures defined in your view model 🤔)
- Is Spotlight supported? I haven't searched if any non-CoreData database could support it 👀

nesium:
Sure, TCA plays nice with UIKit. Obviously you have to take care of navigation yourself then.

In case I have a heavy model (a SQLite table with many columns) and want to present f.e. a list/detail view, I'd create another struct which only contains/fetches the properties needed for displaying a row in the list view, then load the remaining data once the row was selected and is presented in the detail view. So I'd have the same kind of duplication there.

It seems that you absolutely can [index arbitrary data in Spotlight](https://developer.apple.com/documentation/corespotlight/cssearchableitem), although I have only used NSUserActivity so far which can be indexed by Spotlight as well.

RemiBardon:
> It seems that you absolutely can index arbitrary data in Spotlight, although I have only used NSUserActivity so far which can be indexed by Spotlight as well.

Oh, good to know 👍🏻

PR #10, opened 2022-06-02
Implement sidebar "Spotlight" section

### LINKED ISSUE #8: Giving away SwiftUI code
I just pulled the project to check something, and when rebasing my branch against `master`, I realized there are changes I never pushed here. Because of recent modularization (#6), those changes cannot be merged as is. I'm too busy right now, and it'll be too late in June, that's why I'm not making a PR.

@nesium Maybe you want to have a look at them, there is a lot of SwiftUI code you could avoid rewriting.

The full diff can be found here: https://github.com/prose-im/prose-app-macos/compare/79cc1c03446f63a4557dad96c0beae8ffa4be421...RemiBardon:3f95129830a349474bb327364c83ad0f0507d75a

RemiBardon:
Closes #8


Prose/ProseLib/Sources/SidebarFeature/Icon.swift
```diff
@@ -0,0 +1,31 @@
...
+    case directMessage = "message"
+    case addressBook = "text.book.closed"
+    case group = "circle.grid.2x2"
+
+    var image: Image {
+        Image(systemName: self.rawValue)
+    }
+}
+
+struct Icon_Previews: PreviewProvider {
```
nesium, line 22: Hm, do you think we actually need a preview for a small subset of SF Symbols? This also kind of goes for the actual Enum. Do we define a different Enum in every place where we want to use SF Symbols? What is the benefit in this case since there is virtually no code reuse and the strings have to be hard-coded somewhere?

Prose/ProseLib/Sources/ProseUI/Common/SwadowedButtonStyle.swift
```diff
@@ -0,0 +1,47 @@
+//
+//  SwadowedButtonStyle.swift
+//  Prose
+//
+//  Created by Rémi Bardon on 27/03/2022.
+//
+
+import SwiftUI
+
+public struct SwadowedButtonStyle: ButtonStyle {
```
nesium, line 10: The filename/struct name has a typo in it. Should be "ShadowedButtonStyle".

Prose/ProseLib/Sources/SidebarFeature/Icon.swift
RemiBardon: For now, I'm copy-pasting the code I wrote, then I'll translate it to TCA and check that everyting is OK.

nesium (approved):

Prose/ProseLib/Sources/SidebarFeature/Icon.swift
nesium: To solve that problem, I think I'd prefer a snapshot test instead that can be automated. Previews still need to be looked at manually to confirm that everything is working as expected.

nesium (approved):

RemiBardon:
I let you merge if you want, or should I?

valeriansaliou:
Yes, I also think it's up to the author, while others approve.

PR #12, opened 2022-06-02
Fix chat display by using `safeAreaInset`

RemiBardon:
Because of `HStack`s and `VStack`s, the chat screen wasn't interpreted correctly by SwiftUI (scroll under navigation bar and message bar). I fixed it by using `safeAreaInset`, like in the sidebar.

Also, I left

```swift
        // TODO: [Rémi Bardon] Maybe add a material background here, to make it more beautiful with content going under
//        .background(.ultraThinMaterial)
        .background(.background)
```

unchanged (in `MessageBar`), but I really think we should change the background, as it looks a lot better with a material background.

### DIFF
--- a/Prose/ProseLib/Sources/ConversationFeature/Chat/MessageBar/MessageBar.swift
+++ b/Prose/ProseLib/Sources/ConversationFeature/Chat/MessageBar/MessageBar.swift
@@ -44,6 +44,9 @@ struct MessageBar: View {
         // TODO: [Rémi Bardon] Maybe add a material background here, to make it more beautiful with content going under
 //        .background(.ultraThinMaterial)
         .background(.background)
+        // Make sure accessibility frame is correct
+        .contentShape(Rectangle())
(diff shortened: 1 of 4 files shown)


nesium (approved):
Would be nice if you could include two screenshots in such cases in the future. Showing the affected part before and after the change. Because I have no clue what this is about TBH.

RemiBardon:
I thought about it, but right now it doesn't change mush visually (because of opaque bars). When switching to a material message bar, you can see the content scrolling under. I will do it next time.

Anyway, I was just using a tool provided by SwiftUI, to improve accessibility for example.

RemiBardon:
@nesium For example, here you can see the navigation bar having a subtle shadow and material background when the conversation is scrolled. With the previous implementation, SwiftUI had no idea the bar should react to scrolling, and stayed translucent with no border (which was weird).

Screenshot 2022-06-03 at 12 33 46

nesium:
I see. Great attention to detail!

PR #13, opened 2022-06-02
Improve chat info pane

RemiBardon:
I wanted to make the "Info" button work (showing/hiding the info pane), but then I realized I need the whole chat screen to implement TCA… so this makes a lot of changes.

I split them in tiny commits, to help understanding.

I'll fix localization in another PR.


RemiBardon:
I haven't fixed the localisation, but here is the interactive info pane:
Screenshot 2022-06-02 at 19 05 22
Screenshot 2022-06-02 at 19 15 01
Screenshot 2022-06-02 at 19 05 35
It animates on open/close, and fakes loading with a placeholder.

PR #18, opened 2022-06-09
Move stores to TCA environment

RemiBardon:
I also moved this to its own PR

### DIFF
--- a/Prose/ProseLib/Sources/SidebarFeature/SidebarModel.swift
+++ b/Prose/ProseLib/Sources/SidebarFeature/SidebarModel.swift
@@ -25,35 +25,11 @@ public enum SidebarRoute: Hashable {
             hasher.combine(2)
         case .peopleAndGroups:
             hasher.combine(3)
-        case .chat:
+        case let .chat(state):
             hasher.combine(10)
+            hasher.combine(state.chatId)
         case .newMessage:
             hasher.combine(20)
-        }
-    }
-
-    public static func caseEqual(lhs: Self, rhs: Self) -> Bool {
-        switch (lhs, rhs) {
-        case (.unread, .unread),
-             (.replies, .replies),
-             (.directMessages, .directMessages),
-             (.peopleAndGroups, .peopleAndGroups),
-             (.chat, .chat),
-             (.newMessage, .newMessage):
-            return true
-        default:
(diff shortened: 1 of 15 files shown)


Prose/ProseLib/Sources/SidebarFeature/SidebarModel.swift
```diff
@@ -6,16 +6,60 @@
...
+        case .peopleAndGroups:
+            hasher.combine(3)
+        case .chat:
+            hasher.combine(10)
+        case .newMessage:
+            hasher.combine(20)
+        }
+    }
+
+    public static func caseEqual(lhs: Self, rhs: Self) -> Bool {
```
nesium, line 35: Where is that used?

nesium (approved):
Perfect size!

nesium (approved):

Prose/ProseLib/Sources/SidebarFeature/SidebarModel.swift
RemiBardon: By the way, about your question on the difference between the two types of routes in the sidebar, the main issue is this one (I’m having the same problem as yesterday):
In `SidebarContentState`, there is
```swift
public init(
        route: SidebarRoute? = .unread(.init()),
```
In `SpotlightSectionState`, we define the tags of each row
```swift
public struct SpotlightSectionState: Equatable {
    let items: [SidebarItem] = [
        .init(id: .unread(.init()), title: l10n.unreadStack, image: Icon.unread.rawValue, count: 0),
        .init(id: .replies, title: l10n.replies, image: Icon.reply.rawValue, count: 5),
        .init(id: .directMessages, title: l10n.directMessages, image: Icon.directMessage.rawValue, count: 0),
        .init(id: .peopleAndGroups, title: l10n.peopleAndGroups, image: Icon.group.rawValue, count: 2),
    ]
```
When the app starts, it works, since `.unread(.init()) == .unread(.init())`.
However, when we start mutating the state, route changes, but not the SidebarItem.id.
Selecting the row doesn’t work anymore.
That’s why I created the custom `hash(into:)` implementation (which is wrong because it should match the `==` implementation)

nesium (approved):

PR #19, opened 2022-06-09
Migrate `UnreadScreen` to TCA

RemiBardon:


RemiBardon:
OK, one more step done (so many dependencies… that's why my initial PR was huge)

RemiBardon:
@nesium Could you have a quick look at this one first? This way I can merge the other 2 approved stacked PRs 🙂

Prose/ProseLib/Sources/UnreadFeature/UnreadScreen.swift
```diff
@@ -61,68 +56,138 @@ public struct UnreadScreen: View {
...
+                            content: "🆘 socket-1.sgp.atlas.net.crisp.chat - Got HTTP status: \"503 or invalid body\"",
+                            timestamp: Date() - 100_000
+                        ),
+                    ]
+                ),
+            ]))
+            content(state: .init(messages: .init()))
+        }
+
+        @ViewBuilder
```
nesium, line 184: AFAIK `@ViewBuilder` is redundant as it only applies to closures.

nesium (approved):

Prose/ProseLib/Sources/UnreadFeature/UnreadScreen.swift
RemiBardon: Here, it is, as the result is a simple `View`. However, you can put it here if you have conditional views inside (`if`, `switch`), and Swift will understand it's a view and not throw an error saying "Your function has multiple result types".

I put this out of habit, but also keep it because _maybe_ SwiftUI does some optimisations, but on that I have no idea.

nesium: Read/watched more on this. You're absolutely right that it not only applies to closures, as I've incorrectly implied, but wraps the function in a `ViewBuilder` resultBuilder which would allow conditional blocks.

I couldn't find _any_ indication that it would improve anything when a single view is returned.

As such I wouldn't add them explicitly when not needed, but I guess they will keep popping up here and there. There also doesn't seem to be a swiftformat or swiftlint rule for that.

RemiBardon: Oh, thanks for the research 🙏🏻

I was sure it worked for conditional views, as it's the way to avoid wrapping the condition in a `Group`.

I won't put other ones from now (unless I have to), and remove the ones I already put soon.

PR #22, opened 2022-06-09
Add support for multiple windows

RemiBardon:


nesium (commented):
Well structured PR! Revolves around a single topic.

Alas, adding NSWindows to the AppState binds the Reducer and State to a platform and makes it impossible to unit test (without using a TestHost).

Reference types should not go into the state unless there is a very good reason to do so. Importing one of the UI frameworks in a Reducer is usually a sign that something went wrong.

Can we have a view that switches over the route and manages the windows? Like the SwitchStore does for subviews.

The NSWorkspace should also be hidden behind some "…Client", e.g. `WorkspaceClient`. There are examples of that kind in the pointfree isowords repo. Side-effects should be transparent to the reducer and should be injected.

Instead of calling DispatchQueue.main.async you should use the `receive(on:)` method, where the `mainQueue` would be injected from the Environment. This gives you more control in tests by being able to use a `TestScheduler`.

Furthermore I'd suggest that you write a test for the AppReducer which tests all actions and asserts the change of state by using a TestStore. This should give you a better feeling for the distinction between state, side-effects and UI.

RemiBardon:
> Well structured PR! Revolves around a single topic.

Thank you, I tried to do better than the last one 🥲

> Alas, adding NSWindows to the AppState binds the Reducer and State to a platform and makes it impossible to unit test (without using a TestHost).
>
> Reference types should not go into the state unless there is a very good reason to do so. Importing one of the UI frameworks in a Reducer is usually a sign that something went wrong.

Yes, you're right, but I don't know where else to put it, as the environment is immutable, and actions would have the same issues 🤔 Should I keep it outside of TCA scope (i.e. in SwiftUI `.onChange(of:perform:)` for example?

> Can we have a view that switches over the route and manages the windows? Like the SwitchStore does for subviews.

I tried, but `SwitchStore` relies on code internal to TCA, `IfLetStore` uses `_ConditionalContent` which has no equivalent for `Scene`s, and although `WithViewStore` takes the type of its content, we loose the reducer inside its closure. The only solution I see is:

```swift
WindowGroup {
    WithViewStore(self.store.scope(state: /AppRoute.auth)) { viewStore in
        if let state = viewStore.state {
            AuthenticationScreen(store: self.store.scope(
                state: { _ in state },
                action: AppAction.auth
            ))
        }
    }
}
WindowGroup {
    WithViewStore(self.store.scope(state: /AppRoute.main)) { viewStore in
        if let state = viewStore.state {
            MainScreen(store: self.store.scope(
                state: { _ in state },
                action: AppAction.main
            ))
        }
    }
}
```

but I'm not fan of the `{ _ in state }` and it doesn't fix the need to open/close windows manually.

> The NSWorkspace should also be hidden behind some "…Client", e.g. `WorkspaceClient`. There are examples of that kind in the pointfree isowords repo. Side-effects should be transparent to the reducer and should be injected.
>
> Instead of calling DispatchQueue.main.async you should use the `receive(on:)` method, where the `mainQueue` would be injected from the Environment. This gives you more control in tests by being able to use a `TestScheduler`.

Yes, I just want to write PoCs and have things working rapidly, so we can iterate on different things at the same time later.

> Furthermore I'd suggest that you write a test for the AppReducer which tests all actions and asserts the change of state by using a TestStore. This should give you a better feeling for the distinction between state, side-effects and UI.

You're right this would help, but the disctinction is already very clear, I'm just battling with SwiftUI 😥

nesium:
This happens when I submit the login form…

https://user-images.githubusercontent.com/39174/173538307-d3701f0f-7bb0-4405-a8ed-d66147ff43c6.mov

Please put the window handling in the UI layer. We're don't have references to Views in the data layer either. There must be a way. This looks promising: https://developer.apple.com/forums/thread/651592?answerId=680951022#680951022

The call to the workspace still needs to go into the environment.

RemiBardon:
> This happens when I submit the login form…

Yes, I don't know why, but for some reason the Info.plist file wasn't committed. I thought I had fixed it, but apparently not…

RemiBardon:
@nesium Have you pulled the branch before your test? Because it's correctly pushed https://github.com/prose-im/prose-app-macos/pull/22/files#diff-cbee4fd998c5ebfa34d8d6c77b8ac7deb39cb8aba72f78ecda2f8884e40c0170 🤔

nesium:
Yes. I'm seeing the Info.plist changes. Still even after deleting derived data same result.

RemiBardon:
> Yes. I'm seeing the Info.plist changes. Still even after deleting derived data same result.

Same issue here…

nesium:
> Yes. I'm seeing the Info.plist changes. Still even after deleting derived data same result.

The Info.plist is not referenced in the project file (`INFOPLIST_FILE`).

> Ar you thinking about something like

You could call it `openURL`.

RemiBardon:
Yeah, I'm fixing the Info.plist issue, I knew where it came from, I just thought I had already fixed it

RemiBardon:
The thing is that as we don't have a proper development team for the team, I'm never comitting the `.pbxproj` file, and this change was forgotten.

RemiBardon:
> You could call it `openURL`.

Isn't it at the same time too large and too tied to how the window opening logic currently works? I was thinking about something more specific so we can change it later without affecting the calling place.

nesium:
> > You could call it `openURL`.
> 
> Isn't it at the same time too large and too tied to how the window opening logic currently works? I was thinking about something more specific so we can change it later without affecting the calling place.

IIRC windows will be opened via the SwiftUI environment so that call would be removed from the reducer. `openURL` is still something we'll need going forward.

nesium:
> Oh, I committed the `Info.plist` here https://github.com/prose-im/prose-app-macos/pull/23/files#diff-c811e370fd2a30514e3983534110d5689cfe11ff9437ef5fc03367aa02d60920

Use [worktrees](https://andrewlock.net/working-on-two-git-branches-at-once-with-git-worktree/)?

RemiBardon:
> Use [worktrees](https://andrewlock.net/working-on-two-git-branches-at-once-with-git-worktree/)?

Oh, thanks, I'll have a look at that

Prose/Prose/ProseApp.swift
```diff
@@ -6,28 +6,11 @@
...
     var body: some Scene {
-        WindowGroup {
-            AppView()
-        }
-        .windowStyle(DefaultWindowStyle())
-        .windowToolbarStyle(UnifiedWindowToolbarStyle())
-        .commands {
-            SidebarCommands()
-
-            CommandMenu("TODO Menu") {
```
nesium, line 23: Please remove this to re-enable the import Cmd-H shortcut. No need to have todos here.

Prose/ProseLib/Sources/App/AppReducer.swift
```diff
@@ -24,6 +28,36 @@ public let appReducer: Reducer<
...
+                sidebar: .init(credentials: .init(jid: jid))
+            ))
+
+        case let .auth(.loginResult(.success(jid))):
+            if let url = URL(string: "prose://main") {
+                return environment.openURL(url, .init())
+                    .receive(on: environment.mainQueue)
+                    .map { AppAction.didLogIn(jid: jid) }
+                    .mapError(EquatableError.init)
+                    .catch { Just(AppAction.showError($0)) }
```
nesium, line 44: Instead of raising the error to the user, let's crash here. This would be a programmer error and the app be unusable if that path wouldn't work.

Prose/ProseLib/Sources/App/AppReducer.swift
```diff
@@ -24,6 +28,36 @@ public let appReducer: Reducer<
...
     ),
+    Reducer { state, action, environment in
+        switch action {
+        case let .didLogIn(jid):
+            state.route = .main(MainScreenState(
+                sidebar: .init(credentials: .init(jid: jid))
+            ))
+
+        case let .auth(.loginResult(.success(jid))):
+            if let url = URL(string: "prose://main") {
```
nesium, line 39: No need to if/let here. Let's crash (see below for more info).

Here's a helper to isolate the force-unwrapping:

```swift
public extension Optional {
  /// Either returns the unwrapped value or prints `message` and stops execution of the program.
  func expect(_ message: String) -> Wrapped {
    guard let value = self else {
      fatalError(message)
    }
    return value
  }
}
```

Prose/ProseLib/Sources/App/AppScene.swift
```diff
@@ -0,0 +1,113 @@
...
+    public var body: some Scene {
+        WithViewStore(self.store) { viewStore in
+            login()
+                .handlesExternalEvents(matching: Set(arrayLiteral: "login"))
+            main()
+                .handlesExternalEvents(matching: Set(arrayLiteral: "main"))
+                // NOTE: [Rémi Bardon] Make sure `.onChange` is called only once.
+                //       I put it here, but it could be anywhere.
+                .onChange(of: viewStore.route) { newRoute in
+                    // TODO: [Rémi Bardon] Find a way to avoid having this callback called on every change.
```
nesium, line 46: Please no todos "Find a way sometime".

Extend AppRoute like so:

```swift
extension AppRoute {
  enum Tag {
    case auth
    case main
  }
  
  var tag: Tag {
    switch self {
    case .auth:
      return .auth
    case .main:
      return .main
    }
  }
}
```

Then scope the ViewStore to `self.store.scope(state: \.route.tag)` and observe `.onChange(of: viewStore.state)`.

Prose/ProseLib/Sources/App/AppScene.swift
```diff
@@ -0,0 +1,113 @@
...
+            IfLetStore(self.store.scope(
+                state: (\AppState.route).case(/AppRoute.main),
+                action: AppAction.main
+            )) { store in
+                MainScreen(store: store)
+            }
+            .frame(minWidth: 1_280, minHeight: 720)
+            .background(WindowAccessor(window: $mainWindow))
+            .onChange(of: authWindow) { newValue in
+                if let window = newValue {
```
nesium, line 69: Please remove debugging code.

Prose/ProseLib/Sources/App/AppScene.swift
```diff
@@ -0,0 +1,113 @@
...
+                    print("Disposed of auth window.")
+                }
+            }
+        }
+        .windowStyle(DefaultWindowStyle())
+        .windowToolbarStyle(UnifiedWindowToolbarStyle())
+        .commands {
+            SidebarCommands()
+
+            CommandMenu("TODO Menu") {
```
nesium, line 81: Same as above (Cmd-H).

nesium (commented):
Almost there!

Prose/Prose/ProseApp.swift
RemiBardon: I don't understand, do you want me to add id or remove it? Me I didn't touch this, it's been here for ages.

Prose/ProseLib/Sources/App/AppReducer.swift
RemiBardon: I usually have a `URL.init(staticString:) → URL` initializer, but I'll add your more general purpose function.

Prose/ProseLib/Sources/App/AppReducer.swift
RemiBardon: Ok, fair enough

Prose/ProseLib/Sources/App/AppReducer.swift
nesium: That sounds even better actually (staticString).

Prose/ProseLib/Sources/App/AppScene.swift
RemiBardon: Oh, so this time I scope the `ViewStore`? I'm not sure I completely get when you'd like me to scope them or not 🙈

Prose/Prose/ProseApp.swift
nesium: Yes, sorry, you just moved it. Let's remove it though so that we can hide windows via the system shortcut.

Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationScreen.swift
```diff
@@ -0,0 +1,61 @@
...
+//  AuthenticationView.swift
+//  Prose
+//
+//  Created by Marc Bauer on 01/04/2022.
+//
+
+import ComposableArchitecture
+import SwiftUI
+
+public struct AuthenticationScreen: View {
```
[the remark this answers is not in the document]
RemiBardon: Well, it's not an `App`, it's not a `Scene` and it's a `View`, so it could be called `XxxView`. However, I like to distinguish views (understand small pieces of UI) from complete screens. I think it helps understanding what the view is in the big picture.

And Apple does not name every view with the `View` suffix, so I don't find it shocking. But that's up to you 🤷🏻‍♂️

Prose/ProseLib/Sources/App/AppScene.swift
nesium: True, adding and observing `tag` would suffice. Scoping here could still make sense because you have the whole AppState at your hands whereas in the AuthenticationView you only had the values of the form.

Prose/ProseLib/Sources/App/AppReducer.swift
RemiBardon: So are you OK with this?

```swift
import Foundation

extension URL {
	/// Comes from [Constructing URLs in Swift](https://www.swiftbysundell.com/articles/constructing-urls-in-swift/#static-urls)
	public init(staticString string: StaticString) {
		guard let url = URL(string: "\(string)") else {
			preconditionFailure("Invalid static URL string: \(string)")
		}
		self = url
	}
}
```

nesium: Yes, that's great!

Prose/ProseLib/Sources/App/AppScene.swift
RemiBardon: If you say I can scope then I will 😜 As you say, especially when I say I have the whole app state in my hands.

nesium: Scope once per screen 👻

Prose/Prose/ProseApp.swift
RemiBardon: Honestly, I don't think the shortcut worked anyway (I was a bit surprised actually, but since I didn't care, I didn't dig).

It's in the bar, and the action works there, but the shortcut itself does nothing.

Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationScreen.swift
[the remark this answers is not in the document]
RemiBardon: Actually I'm realising my terminology comes from iOS "screens", so I get it's a bit ambiguous on macOS.

Prose/Prose/ProseApp.swift
RemiBardon: Ok, maybe I was blind or my French keyboard messes it up 🤷🏻‍♂️

Anyway, that's a problem for later.

Prose/ProseLib/Sources/App/AppReducer.swift
```diff
@@ -36,17 +36,15 @@ public let appReducer: Reducer<
...
+            return environment.openURL(url, .init())
+                .receive(on: environment.mainQueue)
+                .map { AppAction.didLogIn(jid: jid) }
+                .mapError(EquatableError.init)
+                .catch { error -> AnyPublisher<AppAction, Never> in
+                    fatalError("Failed to open URL <\(url.absoluteString)>: \(error.localizedDescription)")
+                }
+                .eraseToEffect()
 
         case let .showError(error):
```
nesium, line 49: This is now unused.

Prose/ProseLib/Sources/App/AppScene.swift
```diff
@@ -0,0 +1,90 @@
...
+        ))
+    }
+
+    public var body: some Scene {
+        WithViewStore(self.store.scope(state: \.route.tag)) { viewStore in
+            login()
+                .handlesExternalEvents(matching: Set(arrayLiteral: "login"))
+            main()
+                .handlesExternalEvents(matching: Set(arrayLiteral: "main"))
+                // NOTE: [Rémi Bardon] Make sure `.onChange` is called only once.
```
nesium, line 43: Isn't that satisfied now?

nesium (approved):
Two minor things left, otherwise: Well done!

Prose/ProseLib/Sources/App/AppReducer.swift
RemiBardon: What do you mean?

Prose/ProseLib/Sources/App/AppScene.swift
RemiBardon: I wasn't clear

I meant "I put it on `main()`, even though it relates to `auth()` too, but if you put it on `Group { auth(); main() } then it'll be run twice`"

Prose/ProseLib/Sources/App/AppReducer.swift
nesium: The `showError` enum case is unused. Most likely for other failable action the pattern would be to have an enum with a `Result` type where an alert property would be set when the action failed.

Prose/ProseLib/Sources/App/AppScene.swift
nesium: I see. And is that bad? That should be a no-op?

Which of course begs the question why even scope then, etc. Yet here we are. But calling it twice shouldn't be a bad thing and we can remove the comment and leave this thing as is. Or do you have a different view?

Prose/ProseLib/Sources/App/AppReducer.swift
RemiBardon: Oh, good catch! I'm still not used to GitHub showing the selected line as the last line… I thought you were talking about `.catch`

Prose/ProseLib/Sources/App/AppScene.swift
RemiBardon: I can remove the comment, it just felt a bit weird putting the modifier on only one view. Sometimes SwiftUI doesn't run `.onChange`s for views not on screen, but hopefully an empty `WindowGroup` doesn't behave like `EmptyView`.

> Which of course begs the question why even scope then, etc. Yet here we are.

What do you mean?

nesium: And if you put it on a `Group` as you've suggested? It's run twice then, but that seems ok to me.

RemiBardon: I don't really like it being run twice either… I'd like someting like `.drawingGroup`, which really groups things together instead of keeping them individual views, but I don't think it's possible for `Scene`s.

RemiBardon: Btw, it was a `NOTE`, and not a `TODO`

nesium: A `NOTE` saying "Make sure…" sounds like `TODO` to me 😬.

So why not run it twice making sure that …

> Sometimes SwiftUI doesn't run .onChanges for views not on screen

doesn't happen? What are you trying to prevent?

RemiBardon: I did

```swift
Group {
    login()
        .handlesExternalEvents(matching: Set(arrayLiteral: "login"))
    main()
        .handlesExternalEvents(matching: Set(arrayLiteral: "main"))
}
// NOTE: [Rémi Bardon] This `.onChange` is called multiple times, but at least we're sure it runs.
//       See <https://github.com/prose-im/prose-app-macos/pull/22> for discussion about it.
.onChange(of: viewStore.state) { newRoute in
    switch newRoute {
    case .auth:
        mainWindow?.close()
    case .main:
        authWindow?.close()
    }
}
```

nesium:
Ok, one last thing. I might be a sucker for these things. That being said…

> This `.onChange` is called multiple times, but at least we're sure it runs.

It runs **twice**. So while "multiple times" is technically correct it is still imprecise.

Why I'm harping on this is, if you care somewhat about correctness, you'll take these comments seriously. If I'm seeing that comment later I'll be wondering if that is something that _needs_ to be fixed at some point. "Multiple times" might be a serious issue, so I'd have to revisit.

But in this case _I don't think it needs to be fixed_. It might bother you that it runs twice but we're now back to the efficiency of the SwiftUI diffing algorithm, aren't we?

So is it a problem or can we leave it as is, even in the final product?

RemiBardon:
Yesh, we can leave it as is, I'm perfectly fine with the security aspect of it ("at least we're sure it runs"). I had written "twice", but changed for "multiple times" as comments are usually not updated and become incorrect. When we'll add a third, fourth, fifth… window, we will probably forget to update this comment. "Multiple times" will stay true until we move this comment, and this day we'll read and change it accordingly.

Also, it's technically ran "duplicately" twice, but runs many times throughout the app's lifetime, as a normal `.onChange` modifier.

My sentence is a bit ambiguous, but I thought it was fine.

RemiBardon:
Yeah yeah, I get this, that's why it doesn't bother me 😉

nesium:
Then please don't raise attention to it. Otherwise you'll be commenting all View `body`s how often they run and under which circumstances.

RemiBardon:
OK, I remove it and merge?

nesium:
Please & yes!

RemiBardon:
Finally! 😪

Thank you for this review @nesium

Prose/ProseLib/Sources/App/AppReducer.swift
```diff
@@ -24,6 +28,30 @@ public let appReducer: Reducer<
...
+            return environment.openURL(url, .init())
+                .receive(on: environment.mainQueue)
+                .map { AppAction.didLogIn(jid: jid) }
+                .mapError(EquatableError.init)
+                .catch { error -> AnyPublisher<AppAction, Never> in
+                    fatalError("Failed to open URL <\(url.absoluteString)>: \(error.localizedDescription)")
+                }
+                .eraseToEffect()
+
+        default:
```
nesium, line 49: Unfortunately this slipped through. Please no `default` cases. We want the compiler to help us by checking for exhaustiveness. Could you please fix this in #38?

PR #23, opened 2022-06-10
Implement `AuthenticationScreen`

### LINKED ISSUE #24: `AuthenticationScreen` localization

RemiBardon:
Closes #24 and #25.

### DIFF
--- a/Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationReducer.swift
+++ b/Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationReducer.swift
@@ -103,7 +103,6 @@ public let authenticationReducer = Reducer<
         state.alert = nil
 
     case .loginButtonTapped:
-        print("Log in button tapped")
         return Effect(value: .logIn)
 
     case .logIn:
--- a/Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationReducer.swift
+++ b/Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationReducer.swift
@@ -36,6 +36,10 @@ public struct AuthenticationState: Equatable {
     var isFormValid: Bool { self.isAddressValid && self.isPasswordValid }
     var isAddressValid: Bool { !self.jid.isEmpty }
     var isPasswordValid: Bool { !self.password.isEmpty }
+    var isLogInButtonEnabled: Bool { self.isFormValid }
+    /// The action button is shown either when the form is valid or when the login request is in flight
+    /// (for cancellation).
+    var isActionButtonEnabled: Bool { self.isLogInButtonEnabled || self.isLoading }
 
     public init(
         jid: String = "",
@@ -126,23 +130,11 @@ public let authenticationReducer = Reducer<
         state.popover = popover
 
     case .submitTapped(.address):
-        if state.isAddressValid {
-            state.focusedField = .password
-        } else {
-            state.alert = AlertState(
-                title: TextState("Invalid chat address"),
-                message: TextState("Your chat address must not be empty")
-            )
-        }
+        state.focusedField = .password
 
     case .submitTapped(.password):
-        if state.isPasswordValid {
+        if state.isLogInButtonEnabled {
             return Effect(value: .logIn)
-        } else {
-            state.alert = AlertState(
-                title: TextState("Invalid password"),
-                message: TextState("Password must not be empty")
-            )
         }
 
     case .cancelLogInTapped:
--- a/Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationReducer.swift
+++ b/Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationReducer.swift
@@ -17,26 +17,50 @@ import SharedModels
 // MARK: State
 
 public struct AuthenticationState: Equatable {
+    public enum Field: String, Hashable {
+        case address, password
+    }
+
+    public enum Popover: String, Hashable {
+        case chatAddress, noAccount, passwordLost
+    }
+
     @BindableState var jid: String
     @BindableState var password: String
+    @BindableState var focusedField: Field?
+    @BindableState var popover: Popover?
 
-    var isFormValid: Bool {
-        !self.jid.isEmpty && !self.password.isEmpty
-    }
+    var isLoading: Bool
+    var alert: AlertState<AuthenticationAction>?
+
+    var isFormValid: Bool { self.isAddressValid && self.isPasswordValid }
+    var isAddressValid: Bool { !self.jid.isEmpty }
+    var isPasswordValid: Bool { !self.password.isEmpty }
 
     public init(
         jid: String = "",
-        password: String = ""
+        password: String = "",
+        focusedField: Field? = nil,
+        popover: Popover? = nil,
+        isLoading: Bool = false,
+        alert: AlertState<AuthenticationAction>? = nil
     ) {
         self.jid = jid
         self.password = password
+        self.focusedField = focusedField
+        self.popover = popover
+        self.isLoading = isLoading
+        self.alert = alert
     }
 }
 
 // MARK: Actions
 
 public enum AuthenticationAction: Equatable, BindableAction {
-    case loginButtonTapped
+    case alertDismissed
+    case loginButtonTapped, showPopoverTapped(AuthenticationState.Popover)
+    case submitTapped(AuthenticationState.Field), cancelLogInTapped
+    case logIn
 //    case loginResult(Result<UserCredentials, EquatableError>)
     case loginResult(Result<String, AuthenticationError>)
     case binding(BindingAction<AuthenticationState>)
@@ -44,10 +68,15 @@ public enum AuthenticationAction: Equatable, BindableAction {
 
 // MARK: Environment
 
-public struct AuthenticationEnvironment: Equatable {
+public struct AuthenticationEnvironment {
 //    var login: (String, String, ClientOrigin) -> Effect<Result<UserCredentials, EquatableError>, Never>
+    var mainQueue: AnySchedulerOf<DispatchQueue>
 
-    public init() {}
+    public init(
+        mainQueue: AnySchedulerOf<DispatchQueue>
+    ) {
+        self.mainQueue = mainQueue
+    }
 
 //    public init(login: @escaping (String, String, ClientOrigin)
 //        -> Effect<Result<UserCredentials, EquatableError>, Never>)
@@ -62,30 +91,77 @@ public let authenticationReducer = Reducer<
     AuthenticationState,
     AuthenticationAction,
     AuthenticationEnvironment
-> { state, action, _ in
+> { state, action, environment in
+    struct CancelId: Hashable {}
+
     switch action {
+    case .alertDismissed:
+        state.alert = nil
+
     case .loginButtonTapped:
         print("Log in button tapped")
-        if state.isFormValid {
-            return Effect(value: .loginResult(.success(state.jid)))
-        } else {
-            return Effect(value: .loginResult(.failure(.badCredentials)))
+        return Effect(value: .logIn)
+
+    case .logIn:
+        state.focusedField = nil
+        state.isLoading = true
+
+        let isFormValid = state.isFormValid
+        let jid = state.jid
+        return Effect.task {
+            if isFormValid, !jid.starts(with: "error") {
+                return .loginResult(.success(jid))
+            } else {
+                return .loginResult(.failure(.badCredentials))
+            }
         }
+        .delay(for: .seconds(1), scheduler: environment.mainQueue)
+        .eraseToEffect()
+        .cancellable(id: CancelId())
 //        return environment.login(state.jid, state.password, .proseAppMacOs)
 //            .map(AuthenticationAction.loginResult)
 
+    case let .showPopoverTapped(popover):
+        state.focusedField = nil
+        state.popover = popover
+
+    case .submitTapped(.address):
+        if state.isAddressValid {
+            state.focusedField = .password
+        } else {
+            state.alert = AlertState(
+                title: TextState("Invalid chat address"),
+                message: TextState("Your chat address must not be empty")
+            )
+        }
+
+    case .submitTapped(.password):
+        if state.isPasswordValid {
+            return Effect(value: .logIn)
+        } else {
+            state.alert = AlertState(
+                title: TextState("Invalid password"),
+                message: TextState("Password must not be empty")
+            )
+        }
+
+    case .cancelLogInTapped:
+        state.isLoading = false
+        return Effect.cancel(id: CancelId())
+
     case let .loginResult(.success(jid)):
+//    case let .loginResult(.success(credentials)):
         print("Login success: \(jid)")
+        state.isLoading = false
 
     case let .loginResult(.failure(reason)):
-        print("Login failure: \(reason)")
+        print("Login failure: \(String(reflecting: reason))")
+        state.isLoading = false
 
-//    case let .loginResult(.success(credentials)):
-//        return .none
-//
-//    case let .loginResult(.failure(error)):
-//        state.alert = .init(title: .init(error.localizedDescription))
-//        return .none
+        state.alert = .init(
+            title: TextState("Login failure"),
+            message: TextState(reason.localizedDescription)
+        )
 
     case .binding:
         break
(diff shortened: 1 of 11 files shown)


nesium:
We should limit the width of the window…

Screenshot 2022-06-10 at 17 53 34

Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationScreen.swift
```diff
@@ -15,24 +17,205 @@ public struct AuthenticationScreen: View {
...
+            Text("Welcome!")
+                .font(.system(size: 24, weight: .bold))
+            Text("Sign in to your chat account")
+                .foregroundColor(.secondary)
+                .font(.system(size: 16))
+        }
+    }
+
+    private func fields() -> some View {
+        WithViewStore(self.store, removeDuplicates: { $0.isLoading == $1.isLoading }) { viewStore in
```
nesium, line 61: What is it that you want to prevent by providing a custom `removeDuplicates` handler and nesting ViewStores? It seems like an anti-pattern to me. When I remove all nested ViewStores and the `removeDuplicates` handlers it still works as expected.

Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationReducer.swift
```diff
@@ -62,30 +91,77 @@ public let authenticationReducer = Reducer<
...
 
+    case let .showPopoverTapped(popover):
+        state.focusedField = nil
+        state.popover = popover
+
+    case .submitTapped(.address):
+        if state.isAddressValid {
+            state.focusedField = .password
+        } else {
+            state.alert = AlertState(
```
nesium, line 132: I would deem pressing return to have the same effect as clicking the login button. When the login button is disabled nothing happens, otherwise the login routine runs. This would save quite a bit of code since the tracking of the focus field could be removed and we'd save some strings that need to be translated.

Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationScreen.swift
```diff
@@ -15,24 +17,205 @@ public struct AuthenticationScreen: View {
...
+                removeDuplicates: { $0.popover == $1.popover }
+            ) { viewStore in
+                Button("Lost your password?") { actions.send(.showPopoverTapped(.passwordLost)) }
+                    .popover(
+                        unwrapping: viewStore.binding(\.$popover),
+                        case: /State.Popover.passwordLost,
+                        content: { _ in Self.passwordLostPopover() }
+                    )
+                Divider().frame(height: 18)
+                Button("No account yet?") { actions.send(.showPopoverTapped(.noAccount)) }
```
nesium, line 137: Why don't you finish the localization when you're already building the UI and save us both another PR cycle?

RemiBardon: Ok, I will do that 🙂

Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationScreen.swift
RemiBardon: I don't like having `WithViewStore(self.store)` in forms, as it invalidates the full view rendering every time the user taps on a key. That's why I try to scope as much as possible the `viewStore`s. It's quite verbose, but TCA doesn't make it "magical" by inferring all of this, so I do it by hand.

Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationReducer.swift
nesium: I would push error messages and validation as much into the CoreLib as possible.

IMHO we don't need _any_ validation on the Swift side in the authentication form. Forms that validate while I'm typing are annoying, so that would leave us with a validation on submit. This can be done in the core library which then can throw an error if needed.

Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationScreen.swift
nesium: Can you explain why "rendering" the view on every keystroke would be a bad thing? Isn't that just normal SwiftUI behavior which is efficient in calculating diffs between state changes? How would a vanilla (read: no use of TCA) SwiftUI form look like?

RemiBardon: You're right, but since we can avoid re-rendering, why wouldn't we?

Note: I don't know if SwiftUI is smart enough to only re-render the views which used the properties which changed. I think it's the case, but I have no documentation article which explains it in mind.

RemiBardon: Should I create a very basic example to test what I have in mind?

Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationReducer.swift
RemiBardon: I was still thiking about a validation in the Core library, but I get your point.

I will remove this validation.

nesium:
I don't think that limiting the width would impede support for dynamic type, since longer text can wrap and flow along the vertical axis.

Also I didn't say that I'm not a fan of having a reference to `NSWindow` but having an `NSWindow` in the TCA state.

Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationScreen.swift
nesium: > You're right, but since we can avoid re-rendering, why wouldn't we?

We wouldn't because it clutters up the code and makes us do things a computer is very good at. User input as long as it isn't a continuous gesture like a drag is usually very slow. How fast can a user click or type? Would comparing a handful of values this often be a problem for a modern computer? Would we save energy? Should we spend developer time (& money) on optimizing this problem?

nesium: > Note: I don't know if SwiftUI is smart enough to only re-render the views which used the properties which changed. I think it's the case, but I have no documentation article which explains it in mind.

Here are some:
* https://developer.apple.com/videos/play/wwdc2021/10022/
* https://www.vadimbulavin.com/swiftui-view-lifecycle/
* https://www.hackingwithswift.com/quick-start/swiftui/how-to-find-which-data-change-is-causing-a-swiftui-view-to-update
* Also interesting: All questions and answers from the WWDC 22 lounges: https://midnight-beanie-ccb.notion.site/swiftui-lounge-wwdc22-e20094b91f074398ba395c3fa245e63d (See "Assume that I keep the whole app state in the state object in my app struct?…")

So far I have limited experience with SwiftUI myself, so if you find anything let me know. But please let's not clutter up our code on assumptions. If there's a performance issue we can fix it, but let's not optimize prematurely without actually measuring.

RemiBardon: Thank you for the research 🙏🏻

> Would comparing a handful of values this often be a problem for a modern computer?

The thing is, as soon as a node in your tree hierarchy changes, I'm pretty sure all of the underlying nodes are rerendered (because SwiftUI cannot know if the views will take the same amount of space for example).

> User input as long as it isn't a continuous gesture like a drag is usually very slow. How fast can a user click or type?

Fast enough for developers to have to debounce key strokes, but you're right.

> But please let's not clutter up our code on assumptions. If there's a performance issue we can fix it, but let's not optimize prematurely without actually measuring.

As I told you the other day, it's because I already had the issue that I'm fixing this every time. But if you want I can stop doing it.

By the way, I can't find

> Also interesting: All questions and answers from the WWDC 22 lounges: https://midnight-beanie-ccb.notion.site/swiftui-lounge-wwdc22-e20094b91f074398ba395c3fa245e63d (See "Assume that I keep the whole app state in the state object in my app struct?…")

nesium: > The thing is, as soon as a node in your tree hierarchy changes, I'm pretty sure all of the underlying nodes are rerendered (because SwiftUI cannot know if the views will take the same amount of space for example).

Please define "rendering". When I in a UIKit UITextField, is that "rendering" as well? Also are you talking about the actual visible rendering or about the view structure (the SwiftUI `View` struct) which is the underlying data for a rendered view. Why would an actual (drawn) view (think CALayer) be rerendered when its underlying data (think SwiftUI View) doesn't change?

> Fast enough for developers to have to debounce key strokes, but you're right.

That is out of context. Debounce to _do what_? Fire of a network request or write to the file system? Sure, debounce because it's expensive. Relayout your views? No problem, do it in the next runloop cycle (which is essentially a debounce). Compare two tiny structs?

> As I told you the other day, it's because I already had the issue that I'm fixing this every time. But if you want I can stop doing it.

That was another setup/project, wasn't it? Otherwise, please let me have a look to understand the issue.

> > Also interesting: All questions and answers from the WWDC 22 lounges: https://midnight-beanie-ccb.notion.site/swiftui-lounge-wwdc22-e20094b91f074398ba395c3fa245e63d (See "Assume that I keep the whole app state in the state object in my app struct?…")

It's filed under 'best practices'…

Screenshot 2022-06-13 at 10 41 32

RemiBardon: > Please define "rendering". When I in a UIKit UITextField, is that "rendering" as well? Also are you talking about the actual visible rendering or about the view structure (the SwiftUI View struct) which is the underlying data for a rendered view.

I was talking about visible rendering.

> Why would an actual (drawn) view (think CALayer) be rerendered when its underlying data (think SwiftUI View) doesn't change?

Let's say you have:

```swift
HStack {
  if isDisplayed {
    Color.red
  }
  Text("A very long paragraph")
}
```

When `isDisplayed` is `false`, `Text("A very long paragraph")` is rendered with some available space, depending on its layout priority, and the `intrinsicContentSize`s of the other views.

When `isDisplayed` switches to `true`, the underlying data of `Text` is still exactly the same, SwiftUI has already rendered it, probably cached it, but it **needs** to rerender it (otherwise I'd like to understand this magic).

That's why I said

> as soon as a node in your tree hierarchy changes, I'm pretty sure all of the underlying nodes are rerendered

Do you agree?

> That was another setup/project, wasn't it? Otherwise, please let me have a look to understand the issue.

Yes, it was another project, and I had a map next to a text field.

> It's filed under 'best practices'…

Thank you, I couldn't find it with a search (even in the HTML content).

nesium: > Let's say you have:
> 
> ```swift
> HStack {
>   if isDisplayed {
>     Color.red
>   }
>   Text("A very long paragraph")
> }
> ```
> 
> When `isDisplayed` is `false`, `Text("A very long paragraph")` is rendered with some available space, depending on its layout priority, and the `intrinsicContentSize`s of the other views.
> 
> When `isDisplayed` switches to `true`, the underlying data of `Text` is still exactly the same, SwiftUI has already rendered it, probably cached it, but it **needs** to rerender it (otherwise I'd like to understand this magic).

I don't think that's an appropriate example for what you were doing in the AuthenticationScreen. How about this…

```swift
import SwiftUI

class ViewModel: ObservableObject {
  @Published var outer = "Outer"
  @Published var inner = "Inner"
}

struct ContentView: View {
  @ObservedObject var viewModel = ViewModel()
  
  var body: some View {
    VStack {
      OuterView(text: self.viewModel.outer) {
        InnerView(text: self.viewModel.inner)
      }
      Button("Change outer") {
        self.viewModel.outer += "."
      }
      Button("Change inner") {
        self.viewModel.inner += "."
      }
    }
  }
}

struct OuterView<T: View>: View {
  let text: String
  let content: T
  
  init(text: String, @ViewBuilder content: () -> T) {
    self.text = text
    self.content = content()
  }
  
  var body: some View {
    VStack {
      let _ = Self._printChanges()
      AppKitLabel(text: self.text)
      self.content
    }
  }
}

struct InnerView: View {
  let text: String
  
  var body: some View {
    let _ = Self._printChanges()
    AppKitLabel(text: self.text)
  }
}

struct AppKitLabel: NSViewRepresentable {
  let text: String

  func makeNSView(context: Context) -> NSTextField {
    print("Init", self.text)
    return NSTextField(string: self.text)
  }
  
  func updateNSView(_ nsView: NSTextField, context: Context) {
    print("Update", self.text)
    nsView.stringValue = self.text
  }
}
```

Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationReducer.swift
valeriansaliou: I think that, while it's the role of the app to validate that fields are non-empty, it's the role of the core library to do the rest of the validation work (JID validity, and all other server-backed errors), returned as an ENUM.

On the press return to go to next field, on macOS we could do it as such:

1. Press <return> in the JID field, the app focuses on the next field (password)
2. Press <return> in the password field, the app submits the form (if the form is 100% filled).

Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationScreen.swift
nesium: If you change the `InnerView` and `OuterView` as well so that this resembles more what `WithViewStore` does like this:

```swift
struct ContentView: View {
  @ObservedObject var viewModel = ViewModel()
  
  var body: some View {
    VStack {
      OuterView(viewModel: self.viewModel) {
        InnerView(viewModel: self.viewModel)
      }
      Button("Change outer") {
        self.viewModel.outer += "."
      }
      Button("Change inner") {
        self.viewModel.inner += "."
      }
    }
  }
}

struct OuterView<T: View>: View {
  @ObservedObject var viewModel: ViewModel
  let content: T
  
  init(viewModel: ViewModel, @ViewBuilder content: () -> T) {
    self.viewModel = viewModel
    self.content = content()
  }
  
  var body: some View {
    VStack {
      let _ = Self._printChanges()
      AppKitLabel(text: self.viewModel.outer)
      self.content
    }
  }
}

struct InnerView: View {
  @ObservedObject var viewModel: ViewModel
  
  var body: some View {
    let _ = Self._printChanges()
    AppKitLabel(text: self.viewModel.inner)
  }
}
```

The output looks like this…

Start:

```
OuterView: @self, @identity, _viewModel changed.
Init Outer
Update Outer
InnerView: @self, @identity, _viewModel changed.
Init Inner
Update Inner
```


Tap "Change outer":

```
OuterView: _viewModel changed.
Update Outer.
InnerView: _viewModel changed.
```

Tap "Change inner":

```
OuterView: _viewModel changed.
InnerView: _viewModel changed.
Update Inner..
```

More diffing but same amount of updates. This is why we need to have a properly sized `ViewState` for ViewStores in views that have (many) children, like [here](https://github.com/pointfreeco/swift-composable-architecture/blob/87f388bedbf1ce484e639ace9d296016a7f20812/Examples/TicTacToe/tic-tac-toe/Sources/TwoFactorUIKit/TwoFactorViewController.swift#L33).

nesium: > I'm just wondering now, what happens if you have two `InnerView`s next to the other and you change one of them?

There's only one way to find out…

Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationReducer.swift
```diff
@@ -62,30 +95,64 @@ public let authenticationReducer = Reducer<
     AuthenticationState,
     AuthenticationAction,
     AuthenticationEnvironment
-> { state, action, _ in
+> { state, action, environment in
+    struct CancelId: Hashable {}
```
nesium, line 99: No need to act on that, just as a note:

Maybe you already have something like that, however I've came to love this extension…

```swift
public extension Effect {
  // Source: https://forums.swift.org/t/ifletstore-and-effect-cancellation-on-view-disappear/38272/23
  static func cancel<T>(token: T.Type) -> Effect where T: CaseIterable, T: Hashable {
    .merge(token.allCases.map(Effect.cancel(id:)))
  }
}
```

You'd then define "effect tokens", e.g.

```swift
enum InboxEffectToken: CaseIterable, Hashable {
  case upgradeSubscription
  case requestTrackingAuthorization
}
```

and cancel them all at once…

```swift
case .onDisappear:
  .cancel(token: InboxEffectToken.self)
```

nesium (approved):

Prose/ProseLib/Sources/AuthenticationFeature/AuthenticationReducer.swift
RemiBardon: 😯 That's handy 👍🏻 Thank you 🙏🏻

PR #27, opened 2022-06-10
Implement MFA window

RemiBardon:


Prose/ProseLib/Sources/AppLocalization/Resources/en.lproj/Localizable.strings
```diff
@@ -6,23 +6,33 @@
...
+"authentication.basic_auth.password_lost.popover.content" = "Please open this link to our website\nto [recover your password](https://prose.org).";
+"authentication.basic_auth.password_lost.popover.title" = "Lost your password?";
+
+"authentication.mfa.confirm_button.title" = "Confirm login code";
+"authentication.mfa.header.title" = "One more step…";
+"authentication.mfa.header.subtitle" = "Enter your 6-digit access code";
+"authentication.mfa.error.title" = "Multi factor authentication failure";
+"authentication.mfa.footer.cannot_generate_code.title" = "Cannot generate code?";
+// TODO: [Rémi Bardon] Add text
+"authentication.mfa.footer.cannot_generate_code.content" = "TODO";
```
RemiBardon, line 33: See https://prose-im.slack.com/archives/C03G1PS24B1/p1655139897071139

RemiBardon: Wont fix, see https://prose-im.slack.com/archives/C03G1PS24B1/p1655289987035429

Prose/ProseLib/Sources/AuthenticationFeature/MFA6DigitsView.swift
```diff
@@ -0,0 +1,275 @@
...
+public let mfa6DigitsReducer: Reducer<
+    MFA6DigitsState,
+    MFA6DigitsAction,
+    AuthenticationEnvironment
+> = Reducer { state, action, environment in
+    switch action {
+    case .onAppear, .onGlobalTapGesture:
+        state.focusedField = .textField
+
+    case .mainButtonTapped:
```
nesium, line 150: Nitpick: `submitButtonTapped` would be nicer than `mainButtonTapped` since it also carries what the button does not only its placement in the form. Makes it a little easier to comprehend.

Prose/ProseLib/Sources/AuthenticationFeature/MFA6DigitsView.swift
```diff
@@ -0,0 +1,275 @@
...
+            }
+        }
+        // Hide fake fields in accessibility
+        .accessibilityHidden(true)
+        // Add a real `TextField` backing the interactions and accessibility
+        .background {
+            TextField(l10n.Form.OneTimeCode.placeholder, text: viewStore.binding(\.$code))
+                .focused($focusedField, equals: .textField)
+                .textContentType(.oneTimeCode)
+                .opacity(0)
```
nesium, line 76: That was a cool idea with the hidden textfield.

There's just one thing that bothers me: You're filtering out non-numeric characters which is good. However if the user _does_ type a non-numeric character they land in a state that's non obvious. Suddenly they have to press backspace as many times as they've entered invalid characters, but don't see the result. It would be better if the contents of the hidden textfield would reflect what's on screen. 

Here's something that I've [found](https://stackoverflow.com/a/58736068) that does that. I don't know if there's an easier way in SwiftUI?

Prose/ProseLib/Sources/App/AppReducer.swift
```diff
@@ -35,7 +35,7 @@ public let appReducer: Reducer<
                 sidebar: .init(credentials: .init(jid: jid))
             ))
 
-        case let .auth(.loginResult(.success(jid))):
+        case let .auth(.didLogIn(jid, _)):
```
nesium, line 38: I was just thinking if there could be a login failure and who should handle that. Of course the `authenticationReducer` for all but one case: When the user cancels. The app should quit if you close the authentication window*. Currently you're just ending up with a windowless application if you close it. How can we implement this in SwiftUI?

*We're going to reuse the `AuthenticationFeature` if you log into other accounts. If you're already logged into at least one, the app should not quit of course. This is where it starts to make sense that the parent - in this case the `appReducer` decides how to handle the result.

We could have a struct in our SharedModels: `UserCancelledError`. The enum case would then be `.loginResult(Result<Credentials, UserCancelledError>)`.

nesium (commented):
Nicely done.

Prose/ProseLib/Sources/AuthenticationFeature/MFA6DigitsView.swift
RemiBardon: Yeah, ok. I also wanted to change the button in the login screen, as it is called `loginButton` and it does cancellation too…

Prose/ProseLib/Sources/AuthenticationFeature/MFA6DigitsView.swift
RemiBardon: Yeah, I don't know if there's an easy solution. I just told myself that users would not type letters so we'd be good for 99% of them.

I just realised I didn't set the keyboard layout to "decimal digits", maybe this can help (if not, at least it'll set the correct keyboard on iOS (if `.textContentType(.oneTimeCode)` doesn't do it already))

nesium: We should either let the users type garbage in there or filter it out completely. No hidden states.

Prose/ProseLib/Sources/App/AppRoute.swift
```diff
@@ -9,7 +9,7 @@ import AuthenticationFeature
 import MainWindowFeature
 
 public enum AppRoute: Equatable {
-    case auth(AuthenticationState)
+    case auth(AuthRoute)
```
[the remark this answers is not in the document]
RemiBardon: Ok, I'll fix both of these.

> the AuthenticationFeature decides which route to show and not the AppScene

Can't `authenticationReducer` say `state = .basic` for example?

Prose/ProseLib/Sources/AuthenticationFeature/MFA6DigitsView.swift
RemiBardon: Yeah, I agree

RemiBardon: Oh yeah, why did I filter at all?

Prose/ProseLib/Sources/App/AppRoute.swift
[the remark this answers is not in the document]
nesium: Internally, sure:

```swift
public enum AuthenticationState: Equatable {
    var route = Route.basic(BasicAuthState())
}
```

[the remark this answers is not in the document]
RemiBardon: I mean, if the state is an `enum`, `authenticationReducer` can still change the case, with `state =`.

I thought you were saying otherwise

> the `AuthenticationFeature` decides which route to show and not the AppScene

(By the way, I'm just answering, I will fix this anyway)

Prose/ProseLib/Sources/AuthenticationFeature/MFA6DigitsView.swift
RemiBardon: This seemed interesting, unfortunately, as it's an `Int`, it would not support having an empty field.

https://stackoverflow.com/a/70601917/10967642

nesium: The ViewModifier looks quite nice though: https://stackoverflow.com/a/71822350

RemiBardon: Yesterday I gave a few tries at this, but unfortunately I can't override what the user inputs in the `TextField` 😕 The UI and the `Binding` are in fact two separate things, and even though I filter in the `Binding`, it doesn't filter in the field. I'm thinking about doing someting like filtering and then

```swift
DispatchQueue.main.async {
    realBinding.wrappedValue = filteredInput
}
```

so it overrides on next render.

RemiBardon: @nesium I managed to make it work, using https://stackoverflow.com/a/71093025/10967642.

https://user-images.githubusercontent.com/37386490/174240560-39e22324-0fd7-4e71-a610-ddd03130c7c6.mov

(The numbers don't react yet, I've just disabled them)

RemiBardon: I will also block to 6 characters

RemiBardon: It works 🥳 And everything is handled in the Reducer 🙂

https://user-images.githubusercontent.com/37386490/174247125-d4951cce-be48-413d-a701-4f26d141ed24.mov

Prose/ProseLib/Sources/App/AppReducer.swift
RemiBardon: Do I keep this as we got rid of the secondary window?

nesium: Keep what exactly? The login still needs the functionality to be cancelled. We could just say that if there is no account, we don't show a close/cancel button. If we do, the app should quit when it's tapped. If there is an account the sheet should be dismissed.

RemiBardon: I meant "keep this as is". But you're right, we need such feature. Can we just keep it for after #38?

nesium (approved):

PR #37, opened 2022-06-15
Store JID and password in the Keychain and restore state from it

RemiBardon:


Prose/ProseLib/Sources/App/AppReducer.swift
```diff
@@ -30,21 +30,42 @@ public let appReducer: Reducer<
...
+            guard case .auth = state.route else { return .none }
+
+            guard let jid = environment.userDefaults.loadCurrentAccount() else { return .none }
+
+            do {
+                guard let password = try environment.credentials.loadCredentials(jid) else { return .none }
+                return Effect(value: .auth(.didLogIn(jid: jid, password: password)))
+            } catch {
+                // TODO: Handle this error, user might have denied access to the credentials
+                print("Error when loading credentials: \(error.localizedDescription)")
```
nesium, line 43: There's nothing to do for us here. We could log the error for informational purposes but the user made their decision or we have an error that we can do nothing about other than show the login screen again. No need to handle this error.

Prose/ProseLib/Sources/App/AppReducer.swift
```diff
@@ -30,21 +30,42 @@ public let appReducer: Reducer<
...
+                return environment.openURL(url, .init())
+                    .receive(on: environment.mainQueue)
+                    .map { AppAction.didLogIn(jid: jid) }
+                    .mapError(EquatableError.init)
+                    .catch { error -> AnyPublisher<AppAction, Never> in
+                        fatalError("Failed to open URL <\(url.absoluteString)>: \(error.localizedDescription)")
+                    }
+                    .eraseToEffect()
+            } catch {
+                fatalError("Failed to store credentials for '\(jid)': \(error.localizedDescription)")
```
nesium, line 67: No need to crash here. The authentication was successful. A pity that our user will have to enter their credentials again after a restart but why not let them use the app for now?

Prose/ProseLib/Sources/App/AppScene.swift
```diff
@@ -35,6 +35,7 @@ public struct AppScene: Scene {
     }
 
     public var body: some Scene {
+        let _ = actions.send(.onAppear)
```
nesium, line 38: That should certainly go into the `.onAppear(perform:)` handler.

Prose/ProseLib/Sources/SharedModels/CredentialsClient.swift
```diff
@@ -0,0 +1,91 @@
...
+    public var save: (_ jid: JID, _ password: String) throws -> Void
+    public var deleteCredentials: (_ jid: JID) throws -> Void
+}
+
+private enum KeychainError: Error {
+    case unexpectedPasswordData
+    case unhandledError(status: OSStatus)
+}
+
+public extension CredentialsClient {
```
nesium, line 24: When I added the `SharedModels` package my intention was to bundle [domain models](https://martinfowler.com/eaaCatalog/domainModel.html). The `CredentialsClient` and `UserDefaultsClient` are dependencies however.

In another project I've added a package `Toolbox` put almost all dependencies in there and called it a day. The pointfree people add a package for each dependency + dependency "live". That seemed a bit much to me. Yesterday however I saw a great talk that also proposed having "interface packages" (just the dependency but not the "live" implementation) since it can help with the parallelization of the build. Multiple packages would depend on the interface only which is quick to build and are not blocked by the longer compilation time of the implementation.

Thus I'd say let's bite the bullet and add a package for each dependency. But also put the live implementation in there for now. We can later pull out the implementation if we need to and would only have to change the one site which builds up the `AppEnvironment`.

Prose/ProseLib/Sources/App/AppReducer.swift
RemiBardon: Then we still need to handle the error, by showing the login screen.

Prose/ProseLib/Sources/App/AppReducer.swift
RemiBardon: Erm, I'm not sure the app will be 100% usage, won't we need the credentials at other times?

(But you're right, for the moment I'll remove this `fatalError`)

Prose/ProseLib/Sources/App/AppScene.swift
RemiBardon: No, because I want it to be called only once, and not once per window.

I didn't like it either and with #38 it [is now fixed](https://github.com/prose-im/prose-app-macos/pull/38/commits/1592ec7cd70284823401eed611b64c471987b488#diff-7e699972e390677c42757230305df7bd9805efc806d35b0e57f362e3e99cd06aR46).

Prose/ProseLib/Sources/SharedModels/CredentialsClient.swift
RemiBardon: Yeah, I dind't like putting it in `SharedModels` either, so I'll do this.

So we're talking about a `CredentialsClient` and a `UserDefaultsClient` **targets**? Or are you really talking about Swift **packages**?

Prose/ProseLib/Sources/App/AppReducer.swift
nesium: The route is already set to `.auth` in this code path. If you don't crash the login form will be shown.

Prose/ProseLib/Sources/App/AppReducer.swift
nesium: Can you elaborate?

Prose/ProseLib/Sources/App/AppScene.swift
nesium: Now you're calling it everytime `body` gets invalidated which you have fairly little control over. If you want to be certain, add a `isVisible: Bool` property to the AppState and ignore `.onAppear` actions there if is already true.

Prose/ProseLib/Sources/SharedModels/CredentialsClient.swift
nesium: Sorry, **targets**. You're correct.

Prose/ProseLib/Sources/App/AppScene.swift
RemiBardon: Ok, I will do that but in the other PR (referencing the comment there)

Prose/ProseLib/Sources/App/AppReducer.swift
RemiBardon: I mean, will we need to retreive the credentials at other moments than the app startup, to communicate them to ProseCore in order to start new connections for example? Otherwise, how are those connections authenticated? Are the credentials stored in memory in ProseCore (which wouldn't be surprising)?

Prose/ProseLib/Sources/App/AppReducer.swift
RemiBardon: Oh yes, it is on this branch, I just remembered [changing this in the next PR](https://github.com/prose-im/prose-app-macos/pull/38/commits/1592ec7cd70284823401eed611b64c471987b488#diff-7e699972e390677c42757230305df7bd9805efc806d35b0e57f362e3e99cd06aR21).

Do you mind if I leave it like this?

Prose/ProseLib/Sources/App/AppReducer.swift
nesium: Why would there be more than one connection per account/server?

nesium (approved):

PR #38, opened 2022-06-16
Auth window as sheet

RemiBardon:


RemiBardon:
The PR will stay a draft until I finish fixing review changes in #27.

Prose/ProseLib/Sources/App/AppScene.swift
```diff
@@ -12,65 +12,51 @@ import SettingsFeature
...
+                } else: {
+                    MainScreen(store: Store(
+                        initialState: .placeholder,
+                        reducer: Reducer.empty,
+                        environment: MainScreenEnvironment.placeholder
+                    ))
+                    .redacted(reason: .placeholder)
+                }
+                .frame(minWidth: 1_280, minHeight: 720)
+                .onAppear { actions.send(.onAppear) }
```
RemiBardon, line 46: See https://github.com/prose-im/prose-app-macos/pull/37#discussion_r899876709

Prose/ProseLib/Sources/App/AppReducer.swift
```diff
@@ -20,37 +20,45 @@ public let appReducer: Reducer<
...
-                guard let password = try environment.credentials.loadCredentials(jid) else { return .none }
-                return Effect(value: .auth(.didLogIn(jid: jid, password: password)))
-            } catch {
-                // User might have denied access to the credentials
-                print("Error when loading credentials: \(error.localizedDescription)")
-
-                // TODO: [Rémi Bardon] Add an optional error message to the login screen,
-                //       to show why the login screen appears in case of errors like this.
-                state.route = .auth(.basicAuth(BasicAuthState(jid: jid.jidString)))
+            if let jid = environment.userDefaults.loadCurrentAccount() {
```
nesium, line 39: Since this routine grew in complexity once more, now I'd say let's convert this into a proper side-effect. Where `.onAppear` returns an effect with an output value of `AppAction.authenticationResponse(Result<Credentials, EquatableError>)`.

Prose/ProseLib/Sources/App/AppReducer.swift
```diff
@@ -95,18 +96,19 @@ public let appReducer: Reducer<
 // MARK: State
 
 public struct AppState: Equatable {
-    var route: AppRoute
```
nesium, line 98: There was absolutely no reason to remove `AppRoute`. Quite the opposite.

1. We want to have uniformity in how we handle navigation throughout the app.
2. To do so we want to use the best representation for a particular state. In this case you have introduced the potential for creating two invalid states, either where `main` and `auth` are nil or where they're both non-nil. That was prevented and enforced by the compiler before. [See also](https://en.wikipedia.org/wiki/Algebraic_data_type).

nesium (commented):
I'll continue soon. Have to grab a bite.

Prose/ProseLib/Sources/App/AppReducer.swift
nesium: Hold on. I might be dead wrong here. Need to think about this for a bit.

RemiBardon: I agree with you, the thing is "they're both non-nil" is a valid state.

When a user logs into a new account, we keep the current view as is, not as a placeholder, and we show the auth screen onto it.

I could do

```swift
enum AppRoute {
  case auth(main: MainState?, auth: AuthState)
  case main(MainState)
}
```

but I didn't feel it was worth the clutter in the reducer.

nesium: Ok, that was non-sense. Both routes/states are not mutually exclusive anymore. The main state is always present while  the authentication flow can be presented additionally.

That being said, I believe there's still room for improvement.

I mentioned that the main state is always present. This kind of gives away that `main` shouldn't be optional. Currently you're using it as a marker to let the UI know if it should enable the redaction on the main window **and** if it should create a mock `Store` so that the `mainWindowReducer` and its children cannot do anything. To do so you even had to stub out the environment all the way down. The last part shouldn't usually be this hard, since a child environment is derived from its parent environment. The code currently doesn't work this way, but that's a different problem.

The thing is that you now have logic in your UI and if you'd wanted to unit test that the environment passed to the `mainWindowReducer` actually does nothing that would be very hard.

Therefore my suggestion is to move the logic to the reducer. We can disable the `mainWindowReducer` while still leaving the environment intact by creating a higher-order reducer that does that for us.

But first we need an indicator when to disable it. Easiest solution in the current state of things:

```swift
extension AppState {
  var mainWindowDisabled: Bool {
    self.auth != nil
  }
}
```

Note that this might not be true in the future. I don't know about our desired UX when our user already has an account and logs into another one. Do we have a live main window below the sheet or do we always disable and redact? However that condition will be easy enough to change. 

The `mainWindowDisabled` property would then also be used by the `AppScene` to activate the redaction. Simple logic there and testable by a snapshot test.

The higher-level reducer looks like this:

```swift
extension Reducer where State == AppState, Action == AppAction, Environment == AppEnvironment {
  func disabled(when isDisabled: @escaping (AppState) -> Bool) -> Self {
    Reducer { state, action, environment in
      guard !isDisabled(state) else {
        return .none
      }
      return self(&state, action, environment)
    }
  }
}
```

and would be used like so…

```swift
public let appReducer: Reducer<
    AppState,
    AppAction,
    AppEnvironment
> = Reducer.combine([
    …,
    mainWindowReducer.optional().pullback(
        state: \AppState.main,
        action: /AppAction.main,
        environment: { $0.main }
    ).disabled(when: \.mainWindowDisabled),
    Reducer { state, action, environment in
    …
```

Now you can unit test this by sending actions to the `mainWindowReducer` and assert that its state doesn't change/the environment is not touched.

Don't forget to remove the now unused `AppRoute`.

Prose/ProseLib/Sources/App/AppReducer.swift
RemiBardon: I don't really get what you mean. This is a bit verbose, but I'm not sure we would benefit from adding a new `Action` (if that's what you meant). I already created `. requireAuthentication` to make the process easier to understand.

I could add more comments though.

Prose/ProseLib/Sources/App/AppReducer.swift
RemiBardon: Oh, and by the way I'm well aware of algebraic data types… and I tend to use them too much 🙈

RemiBardon: I didn't see your last message when I sent mine, sorry for the unrelated answer

RemiBardon: Ok, now I understand your proposition, and it's a good idea

RemiBardon: > The last part shouldn't usually be this hard, since a child environment is derived from its parent environment. The code currently doesn't work this way, but that's a different problem.

Yes, I know, but sometimes deriving can be very verbose (that's why SwiftUI introduced `EnvironmentValues`… how I wish we had that for `struct`s 😅)

nesium: Not deriving destroys dependency injection.

Prose/ProseLib/Sources/App/AppReducer.swift
nesium: That would be my proposal…

I cannot say much other than that side-effects are (mostly) wrapped into actual `Effect`s, that it has less nesting levels and feels more idiomatic to me 🤷.

It also does away with the "An "optional" reducer at "App/AppReducer.swift:40" received an action when state was "nil" message because the reducer doesn't send `.auth(.didLogIn())` itself anymore.

I've also added a `Credentials` struct which has the fields `jid` and `password`.

```swift
public let appReducer: Reducer<
    AppState,
    AppAction,
    AppEnvironment
> = Reducer.combine([
    authenticationReducer.optional().pullback(
        state: \AppState.auth,
        action: /AppAction.auth,
        environment: { $0.auth }
    ),
    mainWindowReducer.optional().pullback(
        state: \AppState.main,
        action: /AppAction.main,
        environment: { $0.main }
    ),
    Reducer { state, action, environment in
        func proceedToMainFlow(with credentials: Credentials) {
            state.main = MainScreenState(
                sidebar: .init(
                    credentials: .init(jid: credentials.jid),
                    footer: .init(
                        avatar: .init(
                            // TODO: Use a JID type, to avoid this parsing
                            // TODO: Use an image stored by the user (not a preview asset)
                            avatar: "avatars/\(credentials.jid.node ?? "valerian")"
                        )
                    )
                )
            )
            state.auth = nil
        }
        
        func proceedToLogin() {
          state.auth = .init(route:
            .basicAuth(.init(jid: environment.userDefaults.loadCurrentAccount()?.rawValue ?? "")))
        }
    
        switch action {
        case .onAppear where !state.hasAppearedAtLeastOnce:
          return Effect<Credentials?, EquatableError>.result {
              Result {
                try environment.userDefaults.loadCurrentAccount()
                  .flatMap(environment.credentials.loadCredentials)
              }.mapError(EquatableError.init)
          }
          .catchToEffect()
          .map(AppAction.authenticationResult)
        
        case let .authenticationResult(.success(.some(credentials))):
            proceedToMainFlow(with: credentials)
            return .none
        
        case .authenticationResult(.success(.none)):
          proceedToLogin()
          return .none
        
        case let .authenticationResult(.failure(error)):
          print("Error when loading credentials: \(error.localizedDescription)")
          proceedToLogin()
          return .none
        
        case let .auth(.didLogIn(credentials)):
          proceedToMainFlow(with: credentials)
          return .fireAndForget {
              environment.userDefaults.saveCurrentAccount(credentials.jid)
              try? environment.credentials.save(credentials.jid, credentials.password)
          }
        
        case .onAppear, .auth, .main:
          return .none
    }
  }
])
```

nesium (approved):
Thank you! 

What do you think about the changes in the reducer? Better/worse/Don't care?

RemiBardon:
Sorry, I didn't write my reaction here.

I like it, except

```swift
proceedToMainFlow(with: credentials)
return .fireAndForget { /* ... */ }
```

I'm not a big fan of not handling errors. Sometimes it makes complete sense, but here I find it borderline. I just made a minor change to your code to at least print the error.

I also miss some logging, be we haven't introduced it yet, so it's OK. Like here:

```swift
return Effect<Credentials?, EquatableError>.result {
  Result {
    try environment.userDefaults.loadCurrentAccount()
      .flatMap(environment.credentials.loadCredentials)
  }.mapError(EquatableError.init)
}
.catchToEffect()
.map(AppAction.authenticationResult)
```

The error is never logged.

But once again: it's fine 😉

nesium:
I agree. And as you said we need some logging facilities first. Here's an extension on `Publisher` that I'm using in another project for these situations where you cannot do anything about the error but want to log it so that you can at least debug it if needed.

```swift
public extension Publisher {
  /// Swallows any errors and returns a publisher that does not produce a result.
  func catchAndLogError(_ message: String) -> AnyPublisher<Never, Never>
    where Self.Output == Never
  {
    self.catch { error -> AnyPublisher<Never, Never> in
      log.warn(message)
      log.error(error)
      return Empty().eraseToAnyPublisher()
    }.eraseToAnyPublisher()
  }

  /// Swallows any errors and returns a publisher that emits `replacement` as its only result.
  func replaceAndLogError(_ replacement: Output, message: String) -> AnyPublisher<Output, Never> {
    self.catch { error -> AnyPublisher<Output, Never> in
      log.warn(message)
      log.error(error)
      return Just(replacement).eraseToAnyPublisher()
    }.eraseToAnyPublisher()
  }

  func logError() -> AnyPublisher<Output, Failure> {
    self.handleEvents(
      receiveCompletion: { completion in
        if case let .failure(error) = completion {
          log.error(AppError(error))
        }
      }
    )
    .eraseToAnyPublisher()
  }
}
```

PR #52, opened 2022-06-17
chore: Remove unnecessary `ViewBuilder`s (#26)

### LINKED ISSUE #26: Remove unnecessary `@ViewBuilder`s

RemiBardon:
Closes #26.

There are only 3 `@ViewBuilder`s left, all required.

### DIFF
--- a/Prose/ProseLib/Sources/AddressBookFeature/Toolbar.swift
+++ b/Prose/ProseLib/Sources/AddressBookFeature/Toolbar.swift
@@ -23,7 +23,6 @@ struct Toolbar: ToolbarContent {
         }
     }
 
-    @ViewBuilder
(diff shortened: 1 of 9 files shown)


nesium (approved):
So shiny. Thanks for doing that! As I've mentioned in the earlier PR however I don't think there's a reason to be strict about these. They have their reason, do no harm otherwise but aren't magic either.

RemiBardon:
Yeah, I had just noted we should remove them, and poped it from my TODO queue.

In fact, I prefer being able to seach for `ViewBuilder` now and find only the places where it makes sense 😌

RemiBardon:
Oh… I merged it in the wrong order 😭

PR #56, opened 2022-06-20
refactor: Migrate `MessageBarTextField` to TCA

RemiBardon:


Prose/ProseLib/Sources/ConversationFeature/Chat/TypingIndicator.swift
```diff
@@ -12,10 +12,11 @@ import SwiftUI
 private let l10n = L10n.Content.MessageBar.self
 
 struct TypingIndicator: View {
-    @State var firstName: String
+    var typing: [String]
```
nesium, line 15: Ok, for now but `String` might not cut it later on. For example when multiple are typing we might want to sort in the order when they started. But again, good enough for now.

nesium (approved):

Prose/ProseLib/Sources/ConversationFeature/Chat/TypingIndicator.swift
RemiBardon: Yes, I thought about this the other day, but I wanted to make it simple, it's not the role of this PR 😉

You mean we'll have a `[(JID, String)]`? Or an `OrderedDictionary<JID, String>`?

PR #59, opened 2022-06-21
feat: Implement sending DM

RemiBardon:


Prose/ProseLib/Sources/ConversationFeature/ConversationScreen.swift
```diff
@@ -89,6 +92,27 @@ public let conversationReducer: Reducer<
             // TODO: [Rémi Bardon] We should remember the `showingInfo` setting, to avoid hiding if every time
             state.toolbar = ToolbarState(user: user, showingInfo: false)
 
+        case let .chat(.messageBar(.textField(.send(messageContent)))):
+            let chatId = state.chatId
+            return environment.authenticationClient.requireJID()
+                .receive(on: DispatchQueue.main)
```
RemiBardon, line 98: I forgot this, will add it to the environment

nesium:
Could you please remove the `AuthenticationClient`. This is not a part of this PR. The `ConversationFeature` shouldn't have any sense of authentication. This is all handled up in one of the parent reducers. If the user is not authenticated the views of the `ConversationFeature` won't be visible/enabled. If we'd implement it this way, we'd have to check authentication in each of the different targets/features.

RemiBardon:
> Could you please remove the `AuthenticationClient`.

Yes I will, but I don't know what I can replace it by.

> The ConversationFeature shouldn't have any sense of authentication.

I don't think it currently does, but OK

> This is all handled up in one of the parent reducers. If the user is not authenticated the views of the ConversationFeature won't be visible/enabled.

I still don't understand how you can pass a non-optional JID to `ConversationEnvironment`, so that it can be used to create a `Message`. I'm probably missing something 🤔

nesium:
No need to replace it with anything. As I've mentioned already on Slack, just assume that you have a method 

```swift
func sendMessage(jid: JID, text: String) -> Void
```

in your Environment.

I do think that the `ConversationFeature` has a sense of authentication since it has access to the `AuthenticationManager` and makes use of it in the `conversationReducer`.

RemiBardon:
> No need to replace it with anything. As I've mentioned already on Slack, just assume that you have a method
> 
> ```swift
> func sendMessage(jid: JID, text: String) → Void
> ```
> in your Environment.

I'd love to, but how do I create the Environment then? I really have no idea how to avoid having an optional JID.

nesium:
Sorry, that wasn't entirely clear. The only JID you need is the one of the receiver. It would look more like:

```swift
func sendMessage(to jid: JID, text: String) → Void
```

Think of that method as a curried function where the sender JID is embedded already.

RemiBardon:
Yes, I know it was the receiver JID, but where do I get the sender JID that will be needed to create the Message inside the function? When creating this Environment, I will need to create the function, and therefore I will need a reference to the current user's JID. But how can I avoid having an optional JID at the time of creating this Environment?

RemiBardon:
> Edit: The receiver JID originates from the selection in Sidebar.

It can come from `ConversationScreen` too, which knows `ChatID` (is is a bad practice too?)

nesium:
> Yes, I know it was the receiver JID, but where do I get the sender JID that will be needed to create the Message inside the function? When creating this Environment, I will need to create the function, and therefore I will need a reference to the current user's JID. But how can I avoid having an optional JID at the time of creating this Environment?

Just do this for now…

```swift
public struct ConversationEnvironment {
    let authenticationClient: AuthenticationClient

    let userStore: UserStore
    let messageStore: MessageStore
    let statusStore: StatusStore
    let securityStore: SecurityStore
    
    var sendMessage = { (_ to: JID, _ text: String) in
      print("Sending '\(text)' to \(to)")
    }

    public init(
        authenticationClient: AuthenticationClient,
        userStore: UserStore,
        messageStore: MessageStore,
        statusStore: StatusStore,
        securityStore: SecurityStore
    ) {
        self.authenticationClient = authenticationClient
        self.userStore = userStore
        self.messageStore = messageStore
        self.statusStore = statusStore
        self.securityStore = securityStore
    }
}
```

nesium:
> > Edit: The receiver JID originates from the selection in Sidebar.
> 
> It can come from `ConversationScreen` too, which knows `ChatID` (is is a bad practice too?)

You're probably referring to `ConversationState`. That's fine. The `mainReducer` initializes `ConversationState` by taking the `JID` from the sidebar selection. That's what I meant.

RemiBardon:
> Just do this for now…

Then I should remove `let authenticationClient: AuthenticationClient`, right?

nesium:
> > Just do this for now…
> 
> Then I should remove `let authenticationClient: AuthenticationClient`, right?

Yes, I just copy & pasted the whole thing.

Prose/ProseLib/Sources/App/AppReducer.swift
```diff
@@ -33,11 +33,17 @@ public let appReducer: Reducer<
     ).disabled(when: \.isMainWindowDisabled),
     Reducer { state, action, environment in
         func proceedToMainFlow(with credentials: Credentials) {
-            state.main = MainScreenState(jid: credentials.jid)
+            if state.isMainWindowRedacted {
+                state.main = MainScreenState(jid: credentials.jid)
+            } else {
+                state.main.sidebar.credentials = credentials.jid
+            }
+            state.isMainWindowRedacted = false
```
RemiBardon, line 41: This doesn't make sense anymore, but should I keep it?

Prose/ProseLib/Sources/MainWindowFeature/MainScreen.swift
```diff
@@ -8,11 +8,13 @@
 import AddressBookFeature
 import ComposableArchitecture
 import ConversationFeature
+import ProseCoreStub
 import SharedModels
 import SidebarFeature
 import SwiftUI
 import TcaHelpers
 import UnreadFeature
+import UserDefaultsClient
```
RemiBardon, line 17: Same here

Prose/ProseLib/Sources/ProseCoreStub/MessageStore.swift
```diff
@@ -15,18 +15,30 @@ public struct Message {
...
+        self.senderId = senderId
+        self.content = content
+        self.timestamp = timestamp
+    }
 }
 
 public struct MessageStore {
     public var messages: (_ chatId: ChatID) -> [Message]?
     public var unreadMessages: () -> OrderedDictionary<ChatID, [Message]>
+    public var sendMessage: (_ chatId: ChatID, _ message: Message) -> Void
```
RemiBardon, line 33: This isn't used anymore, should I revert it?

Prose/ProseLib/Sources/App/AppReducer.swift
nesium: If it doesn't make sense, then no.

Prose/ProseLib/Sources/ProseCoreStub/MessageStore.swift
nesium: Sure

nesium (approved):

Issue #45, opened 2022-06-17
Sidebar: Lock some features


valeriansaliou:
* [ ] Grey out the top-left call button
  * Description: Calls will be implemented later.

* [x] Grey out groups in the roster
  * Description: MUCs will be implemented soon.

* [ ] Grey out the spotlight section
  * Description: Will be implemented soon.
  * Timeline: whenever we implement this, here's the order:
    1. People and groups (backed by the server)
    2. Unread stack (backed from local store after receiving pending messages from the server)
    3. DMs (filtered from local store after fetching all recent MAM)
    4. Replies (filtered from local store after fetching all recent MAM)

* [ ] Grey out account switcher
  * Description: Multi-workspace capabilities TBD, as they may require some extra work on the core-client library.

PR #60, opened 2022-06-21
Refactor sidebar

nesium:
Like the title says, I've converted the sidebar into a plain menu and moved states and views for the actual content into the `MainScreen`.

As mentioned on Slack the "Favorites" section is gone until we can create favorites.

Since the code is part of the sidebar and I had to touch individual states for the "placeholder" state, I've removed unneccessary `public` access scopes e.g. from `FooterState` which doesn't need to be known outside of the `SidebarFeature`.

Furthermore I've removed arguments from initializers that also [do not need to be known](https://en.wikipedia.org/wiki/Law_of_Demeter) from the outside. We can still modify properties of these structs, for example in unit tests, after initializing them.

For simplicity I've left out the single "unredactions" since our user cannot interact with the obscured interface anyway.

I've also typealiased `UserCredentials` to `JID` and added the jid to the `FooterState`. This needs to be done properly once we have information about the current user. It would be more like a `UserInfo` or `CurrentUser` struct then.

I've also did some light refactoring of the (before) `ActionRow` and `NavigationRow`. `ActionRow` has become `ActionButton` since unlike a row it is not selectable and `NavigationRow` has become `IconRow` since the Sidebar doesn't do navigation anymore. It also now only accepts an `Icon` for its `icon` parameter.

`Counter` now accepts a non-optional `Int` because that's what we will work with by default and `0` (zero) already represents "nothing".

Finally I've replaced environments with `Void` in the sidebar where they weren't used, moved `SpotlightGroupBoxStyle` to `UnreadFeature` where it is used and got rid of SectionViews/-reducers because the Sidebar code is already very compact now.


RemiBardon:
I'm reading everything 👀

Just very quickly so you can maybe act on it,

> moved `SpotlightGroupBoxStyle` to `UnreadFeature`

`SpotlightGroupBoxStyle` needs to be used in "Unread", "Replies", "Direct messages" and "People & groups" (see artboards), which makes every row of the "Spotlight" section, hence the name.

Prose/ProseLib/Sources/MainWindowFeature/MainScreen.swift
```diff
@@ -40,92 +54,91 @@ public struct MainScreen: View {
...
-    environment: \MainScreenEnvironment.sidebar
-)
+>.combine([
+    sidebarReducer.pullback(
+        state: \.sidebar,
+        action: CasePath(MainScreenAction.sidebar),
+        environment: { _ in }
+    ),
+    unreadReducer._pullback(
+        state: (\MainScreenState.route).case(CasePath(MainScreenState.Route.unreadStack)),
```
RemiBardon, line 68: I see you used `CasePath.init(_:)`, thanks 🙂

Prose/ProseLib/Sources/MainWindowFeature/MainScreen.swift
```diff
@@ -40,92 +54,91 @@ public struct MainScreen: View {
...
+    ),
+    conversationReducer._pullback(
+        state: (\MainScreenState.route).case(CasePath(MainScreenState.Route.chat)),
+        action: CasePath(MainScreenAction.chat),
+        environment: { _ in .stub }
+    ),
+    Reducer { state, action, _ in
+        switch action {
+        case .sidebar(.selection(.unreadStack)):
+            state.route = .unreadStack(.init())
```
RemiBardon, line 80: So we're resetting the `UnreadStackState` every time the user leaves and comes back? Isn't this a bad UX?

I had cached it at some point, what do you suggest?

Prose/ProseLib/Sources/MainWindowFeature/MainScreen.swift
```diff
@@ -40,92 +54,91 @@ public struct MainScreen: View {
...
+        case .sidebar(.selection(.directMessages)):
+            state.route = .directMessages
+            return .none
+
+        case .sidebar(.selection(.peopleAndGroups)):
+            state.route = .peopleAndGroups
+            return .none
+
+        case let .sidebar(.selection(.chat(jid))):
+            state.route = .chat(.init(chatId: .person(id: jid), recipient: "Huh?"))
```
RemiBardon, line 96: What is this "Huh?"?

`recipient` is the display name of the recipient, currently used in the "… is typing" indicator. I agree it's not perfect, but because everything is not well connected with ProseCore yet, I had to make it a simple `String` like this…

nesium:
> I'm reading everything 👀
> 
> Just very quickly so you can maybe act on it,
> 
> > moved `SpotlightGroupBoxStyle` to `UnreadFeature`
> 
> `SpotlightGroupBoxStyle` needs to be used in "Unread", "Replies", "Direct messages" and "People & groups" (see artboards), which makes every row of the "Spotlight" section, hence the name.

I see. I'll move it into `ProseUI` then.

Prose/ProseLib/Sources/SidebarFeature/Atoms/Counter.swift
```diff
@@ -8,14 +8,10 @@
 import SwiftUI
 
 struct Counter: View {
-    private let count: UInt16?
-
-    init(count: UInt16? = nil) {
-        self.count = count
-    }
+    var count = 0
```
RemiBardon, line 11: I don't remember why this was an `Optional`, but it's better now 👍🏼

Prose/ProseLib/Sources/SidebarFeature/Footer/Footer.swift
```diff
@@ -16,7 +16,7 @@ private let l10n = L10n.Sidebar.Footer.self
 
 /// The small bar at the bottom of the left sidebar.
 struct Footer: View {
-    typealias State = (FooterState, UserCredentials)
+    typealias State = FooterState
```
RemiBardon, line 19: Oh yes, no tuple anymore 😍

Prose/ProseLib/Sources/SidebarFeature/Footer/Footer.swift
```diff
@@ -58,40 +58,43 @@ struct Footer: View {
 
 // MARK: Reducer
 
-public let footerReducer: Reducer<
-    (FooterState, UserCredentials),
+let footerReducer: Reducer<
```
RemiBardon, line 61: Good catch for the `public`, I had put it by mistake.

Prose/ProseLib/Sources/SidebarFeature/Footer/Footer.swift
```diff
@@ -58,40 +58,43 @@ struct Footer: View {
...
+    FooterState,
     FooterAction,
     Void
 > = Reducer.combine([
```
RemiBardon, line 65: One day I should rewrite all `: Reducer<...> = Reducer.combine([` to `= Reducer<...>.combine([`, but that's for another time 😌

Prose/ProseLib/Sources/SidebarFeature/Footer/Footer.swift
```diff
@@ -126,12 +129,7 @@ public enum FooterAction: Equatable {
         }
 
         static var previews: some View {
-            let state = (
-                FooterState(
-                    avatar: .init(avatar: .init(url: PreviewAsset.Avatars.valerian.customURL))
-                ),
-                UserCredentials(jid: "valerian@crisp.chat")
-            )
+            let state = FooterState(credentials: "valerian@crisp.chat")
```
RemiBardon, line 132: Couldn't `credentials` be renamed to `jid`?
I know you talked about `UserInfo` maybe, so could this be renamed to `userInfo`?
I find `credentials` being badly named, as it doesn't contain private data (but that's probably just me and my French translation).

Prose/ProseLib/Sources/SidebarFeature/Rows/ActionButton.swift
```diff
@@ -27,9 +26,8 @@ struct ActionRow: View {
 
 struct ActionRow_Previews: PreviewProvider {
     static var previews: some View {
-        ActionRow(
+        ActionButton(
             title: "sidebar_groups_add",
```
RemiBardon, line 30: Oh, some not localized content which I didn't catch before! I'll fix it when I'll do the settings page (which is full of not localized content).

Prose/ProseLib/Sources/SidebarFeature/SidebarReducer.swift
```diff
@@ -0,0 +1,73 @@
+import ComposableArchitecture
+import Foundation
+import SharedModels
+
+public typealias UserCredentials = JID
```
RemiBardon, line 5: As said just before, maybe we can rename the type right now, just so we don't keep around a name closely related to `AuthenticationFeature`'s `UserCredentials`.

WDYT?

Prose/ProseLib/Sources/SidebarFeature/SidebarReducer.swift
```diff
@@ -0,0 +1,73 @@
...
+
+public struct SidebarState: Equatable {
+    var credentials: UserCredentials
+    var selection: Selection?
+    var footer: FooterState
+    var toolbar = ToolbarState()
```
RemiBardon, line 11: I get why you put this here, but one day we'll probably need to start the app composing a new message, and we'll need the "is composing" boollean to be `true`. I guess that's not a problem, and we'll see later. I was just writing as it crossed my mind.

Prose/ProseLib/Sources/SidebarFeature/SidebarView.swift
```diff
@@ -5,179 +5,80 @@
...
 import SwiftUI
 
-// MARK: - View
-
-/// The sidebar on the left of the screen.
 public struct SidebarView: View {
-    public typealias State = SidebarState
-    public typealias Action = SidebarAction
+    public typealias ViewState = SidebarState
+    public typealias ViewAction = SidebarAction
```
RemiBardon, line 14: You talked about specific `ViewState`s today.
Shouldn't we keep the separation between the `typealias` I named `State` for the `Store` and `ViewState` for the `ViewStore`s?

Prose/ProseLib/Sources/SidebarFeature/SidebarView.swift
```diff
@@ -5,179 +5,80 @@
...
-            securityStore: self.securityStore
-        )
+    private var groupsSection: some View {
+        Section(L10n.Sidebar.Groups.title) {
+            IconRow(title: "My group", icon: .group)
+            ActionButton(title: L10n.Sidebar.Groups.Add.label) {
+                self.viewStore.send(.addGroupButtonTapped)
+            }
+        }
     }
```
RemiBardon, line 83: Are we hard-coding those names for now?

Prose/ProseLib/Sources/SidebarFeature/Toolbar.swift
```diff
@@ -81,13 +81,7 @@ public let toolbarReducer: Reducer<
...
 
 public struct ToolbarState: Equatable {
-    @BindableState var writingNewMessage: Bool
-
-    public init(
-        writingNewMessage: Bool = false
-    ) {
-        self.writingNewMessage = writingNewMessage
-    }
+    @BindableState var writingNewMessage = false
```
RemiBardon, line 84: 👍🏼

Prose/ProseLib/Sources/UnreadFeature/SpotlightGroupBoxStyle.swift
```diff
@@ -8,6 +8,10 @@
 import SharedModels
 import SwiftUI
 
+extension GroupBoxStyle where Self == SpotlightGroupBoxStyle {
+    static var spotlight: Self { SpotlightGroupBoxStyle() }
+}
```
RemiBardon: See the main discussion.

As I see "Refresh" on the top of my page, I guess you already fixed this.

RemiBardon: I put the comment here, but it's for the file in fact.

Prose/ProseLib/Sources/UnreadFeature/UnreadScreen.swift
```diff
@@ -32,6 +32,7 @@ public struct UnreadScreen: View {
             .background(Colors.Background.message.color)
             .toolbar(content: Toolbar.init)
             .onAppear { actions.send(.onAppear) }
+            .groupBoxStyle(.spotlight)
```
RemiBardon, line 35: This is just nitpicking, but as I said about the GroupBoxStyle belonging in the spotlight section, I'd find it logical that we apply the style there.

RemiBardon (commented):
That's very good! I share the feedback, but I'll answer again after I try the app.

I suppose you already fixed some of the issues, we can close the related conversations if so.

RemiBardon:
> I see. I'll move it into `ProseUI` then.

Why `ProseUI` and not `MainWindowFeature`? Is it because of https://github.com/prose-im/prose-app-macos/pull/60#discussion_r902782144?

Prose/ProseLib/Sources/MainWindowFeature/MainScreen.swift
nesium: Unread messages are not static. You don't have 3 unread messages between viewing the "Unread Stack". You receive new messages which are unread by default, so just caching the state isn't a solution.

Prose/ProseLib/Sources/SidebarFeature/Footer/Footer.swift
RemiBardon: Me too, but I had no idea how to name it 😆

Prose/ProseLib/Sources/SidebarFeature/Footer/Footer.swift
nesium: I'm not touching this until we have proper data. If you look at the initializer of `FooterState` you'll recognize that there is work to be done.

Prose/ProseLib/Sources/SidebarFeature/Rows/ActionButton.swift
nesium: That's fine in a preview, isn't it?

Prose/ProseLib/Sources/MainWindowFeature/MainScreen.swift
RemiBardon: Ok, then maybe add a TODO comment?

Prose/ProseLib/Sources/SidebarFeature/SidebarReducer.swift
nesium: See above.

Prose/ProseLib/Sources/MainWindowFeature/MainScreen.swift
RemiBardon: Ok, fair enough

Prose/ProseLib/Sources/SidebarFeature/Rows/ActionButton.swift
RemiBardon: Yes, but we should at least write some human-readable text (i.e. "Example"), this was intended to be localized (`LocalizedStringKey`).

Prose/ProseLib/Sources/SidebarFeature/SidebarReducer.swift
nesium: Yes, we'll handle that later. There are more problems than that to enable such a feature.

Prose/ProseLib/Sources/SidebarFeature/Rows/ActionButton.swift
RemiBardon: (But it's priority -1000, I'll just do it if some day I pass by doing some localization)

Prose/ProseLib/Sources/SidebarFeature/SidebarView.swift
nesium: I can remove the typealias if you'd prefer that. I don't think we need 4 typealiases if we have a Store and a ViewStore.

Prose/ProseLib/Sources/SidebarFeature/SidebarView.swift
nesium: Yes. Loading contacts from the Roster will be the next step. Multi-User-Chat and thus groups will still take some time.

Prose/ProseLib/Sources/UnreadFeature/UnreadScreen.swift
nesium: Maybe the style should be renamed then. I don't think that the `MainScreen` should decorate some views and not others. It's better that `MainScreen` doesn't make any assumptions about what the content looks like and just displays it, no?

Prose/ProseLib/Sources/MainWindowFeature/MainScreen.swift
nesium: The feature is yet to be done. If you build it according to the design is there any other way than to change this?

RemiBardon (changes requested):
I can't click on the hard-coded group

nesium:
> I can't click on the hard-coded group

I know. #45 even says it should be greyed out. Why should it be selectable?

Prose/ProseLib/Sources/UnreadFeature/UnreadScreen.swift
RemiBardon: I don't agree, you wrote:

```swift
List(selection: viewStore.binding(
    get: \.selection,
    send: SidebarAction.selection
)) {
    self.spotlightSection
    self.contactsSection
    self.groupsSection
}
.listStyle(.sidebar)
```

Why not

```diff
List(selection: viewStore.binding(
    get: \.selection,
    send: SidebarAction.selection
)) {
    self.spotlightSection
+         .groupBoxStyle(.spotlight)
    self.contactsSection
    self.groupsSection
}
.listStyle(.sidebar)
```

(Or, better, `.groupBoxStyle(.spotlight)` in `self.spotlightSection`, that was for the code snippet)

(And, I don't know why you mention `MainScreen`)

Prose/ProseLib/Sources/SidebarFeature/SidebarView.swift
RemiBardon: I meant why not hard-coding it as a default value in State, so we can override it, but it's useless, you're right

Prose/ProseLib/Sources/UnreadFeature/UnreadScreen.swift
nesium: [It's applied](https://github.com/prose-im/prose-app-macos/blob/988426165be8d566e68014ff41458942ea1fd54a/Prose/ProseLib/Sources/SidebarFeature/NavigationDestinationView.swift#L44) to the `UnreadScreen` at the moment. Are you sure that would be equivalent?

Prose/ProseLib/Sources/SidebarFeature/SidebarView.swift
RemiBardon: I didn't mean to remove them, just keep the same convention as in other files, and I believe `State` for the `Store` and `ViewState` for the `ViewStore`s is quite logical, isn't it?

(We don't need 4, we can have only `ViewState` and `ViewAction`, but as long as they are used for a `ViewStore`)

Prose/ProseLib/Sources/MainWindowFeature/MainScreen.swift
RemiBardon: No no, that's why I said add a TODO comment, so we don't forget it when implementing the feature 😉 But I'm fine if we don't add such a comment 😉

nesium: I generally don't like TODO message in code. Either fix it or create a ticket. In this case we already have a ticket.

Prose/ProseLib/Sources/SidebarFeature/Rows/ActionButton.swift
nesium: Yeah, just fix it next time you look at that preview.

Prose/ProseLib/Sources/SidebarFeature/SidebarView.swift
nesium: Actually I don't agree. There's only ViewState and ViewAction. Unless you create a `ViewState` struct or `ViewAction` enum which you scope your Store to, your store contains the `ViewState` and `ViewAction` because it is the same.

nesium: I give in.

RemiBardon: Makes sense, you're right

nesium: It's not a clear distinction and the names of the generic parameters Store and ViewStore are both the same (State and Action). They're using State and Action as typealiases in the isowords project as well, so I'll follow suite.

RemiBardon (changes requested):

RemiBardon:
Last minor changes and we'll ge good to go!

RemiBardon (approved):
Except the minor bug https://github.com/prose-im/prose-app-macos/pull/60#pullrequestreview-1013865738, we're good for a merge!

Thank you for this refactor 🙏🏼

Prose/ProseLib/Sources/UnreadFeature/UnreadScreen.swift
nesium: I'll leave it in ProseUI everything else doesn't make sense. But at some point it should probably be renamed to _what it is_ not _where it originates from_.

RemiBardon:
Thank you for your last commit, that's perfect 🙏🏼

nesium:
> Thank you for your last commit, that's perfect 🙏🏼

That was a great catch actually. I didn't notice the problem at first.

PR #61, opened 2022-06-22
chore: Fix sidebar footer action menu design

RemiBardon:
The sidebar footer action menu didn't follow the design artboards. I did better with the avatar button, but never took the time to fix the action menu. Here it is.

The artboards show:
Screenshot 2022-06-22 at 13 51 22

The current implementation was:
Screenshot 2022-06-22 at 13 29 41

and the avatar popover was:
Screenshot 2022-06-22 at 13 30 08

I reused the styles I had created to change the new popover to:
Screenshot 2022-06-22 at 13 50 43

The "Switch account" picker cannot follow the spec, I tried hard but SwiftUI plays against me 😕 The new design is better I think.

I also fixed the account icons by the way.

### DIFF
--- a/Prose/ProseLib/Sources/SidebarFeature/Footer/FooterActionMenu.swift
+++ b/Prose/ProseLib/Sources/SidebarFeature/Footer/FooterActionMenu.swift
@@ -57,7 +57,7 @@ struct FooterActionMenu: View {
 
     fileprivate static func popover(store: Store<State, Action>) -> some View {
         let actions: ViewStore<Void, Action> = ViewStore(store.stateless)
-        return VStack(alignment: .leading) {
+        return VStack(alignment: .leading, spacing: 16) {
             // TODO: [Rémi Bardon] Refactor this view out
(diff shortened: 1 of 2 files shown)


nesium (approved):
I liked the way it was better. Now you have more clicks/delays to switch a team.

valeriansaliou:
I think at some point we'll have a sidebar switcher, only if requested by users, but for now I think Prose will be used on very few teams at the same time, as you can basically use your main team's XMPP account to join other communities under a centrally managed identity. I'd stick to the Sketch designs for that, visually I think it was simpler and more macOS-ish feeling. Switching teams is not so much of a deal right now, which is why I made it nested in a sub-menu in the current designs.

RemiBardon:
> I'd stick to the Sketch designs for that, visually I think it was simpler and more macOS-ish feeling.

I like the macOS feeling too, it's just that SwiftUI doesn't allow this, for compatibility with other platforms.

PR #69, opened 2022-06-24
Show unread badge in sidebar

nesium:
- Updated core lib which now supports roster items without groups and puts them into a (not yet localized) default group
- Unread badges are shown in the sidebar as new messages come in


nesium:
I'm sorry that this has grown quite a bit since the initial commits. Most of the code is boilerplate for communicating with FFI and the test setup however.

What's new:
- Unread counts should now been displayed correctly (hopefully)
- User presence is shown in the sidebar (online, offline)
- Local notifications are generated for incoming messages
- Roster items shouldn't disappear anymore (the bug where the app needs to be restarted after login for the roster to load is still in there)

There's new core lib functionality lingering that can be used:
- Observing and sending chat states (composing, active, paused, etc.)
- Sending & receiving presence
- Updating (editing) messages

I've also noticed a bug in the ConversationScreen: When a message is received when no conversation was displayed yet (i.e. the WebView hasn't been loaded) and you select the conversation, the message is not visible even though the reducer has received it. I could imaging that it has something to do with the loading of the WebView but that's just speculation on my part.

valeriansaliou:
On the Web View there’s a possibility there are some bugs in the early implementation I provided Remi with last week.

I’ve made improvements and hard-tested everything over the weekend, as well as stabilized the data models and store mutation API, I’m pretty sure that any issue related to the views internals are now fixed, if you saw issues related to this.

nesium:
I didn't investigate, but maybe it's fixed then.

Prose/Prose.xcodeproj/project.pbxproj
```diff
@@ -18,6 +18,7 @@
 /* End PBXBuildFile section */
 
 /* Begin PBXFileReference section */
+		2C16617F2867F4400041E270 /* AllTests.xctestplan */ = {isa = PBXFileReference; lastKnownFileType = text; path = AllTests.xctestplan; sourceTree = "<group>"; };
```
RemiBardon, line 21: Doesn't it work to have it in the Swift package?

Prose/Prose.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -63,6 +63,15 @@
...
     },
+    {
+      "identity" : "swift-tagged",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/pointfreeco/swift-tagged",
+      "state" : {
+        "revision" : "f3f773a5e13f3c8f0ab1ce2ae6378058acefa87d",
+        "version" : "0.7.0"
+      }
+    },
```
RemiBardon, line 74: I didn't know this library!
(It's only for prefixed types but I created https://github.com/RemiBardon/swift-prefixed-type, if one day you need this kind of stuff)

Prose/ProseLib/Package.swift
```diff
@@ -31,7 +31,8 @@ let package = Package(
             url: "https://github.com/pointfreeco/swiftui-navigation",
             .upToNextMajor(from: "0.1.0")
         ),
-        .proseCore("0.1.4"),
+        .package(url: "https://github.com/pointfreeco/swift-tagged", .upToNextMajor(from: "0.7.0")),
+        .proseCore("0.1.6"),
```
RemiBardon, line 35: 😯

Prose/ProseLib/Sources/App/AppReducer.swift
```diff
@@ -47,14 +50,26 @@ public let appReducer: Reducer<
...
+                        try environment.userDefaults.loadCurrentAccount()
+                            .flatMap(environment.credentials.loadCredentials)
+                    }.mapError(EquatableError.init)
+                }
+                .catchToEffect()
+                .map(AppAction.authenticationResult),
+
+                .fireAndForget {
+                    environment.notifications.promptForPushNotifications()
+                },
```
RemiBardon, line 66: Maybe we could do this when the user starts their first conversation?
It's always annoying to have to answer all those popups when launching an app for the first time…

Prose/ProseLib/Sources/NotificationsClient/NotificationsClient+Live.swift
```diff
@@ -0,0 +1,76 @@
...
+    static let live: Self = {
+        let permissionSubject = CurrentValueSubject<NotificationPermission, Never>(.notDetermined)
+
+        func refreshNotificationPermission() {
+            UNUserNotificationCenter.current().getNotificationSettings { settings in
+                permissionSubject.send(.init(authorizationStatus: settings.authorizationStatus))
+            }
+        }
+
+        refreshNotificationPermission()
```
RemiBardon, line 17: This is smart. I just wonder if it could be delayed without adding too much complexity 🤔

Prose/ProseLib/Sources/ProseCoreTCA/Types/ChatState.swift
```diff
@@ -0,0 +1,60 @@
...
+    public var timestamp: Date
+
+    public init(kind: ChatState.Kind, timestamp: Date) {
+        self.kind = kind
+        self.timestamp = timestamp
+    }
+}
+
+public extension ChatState {
+    /// See: https://xmpp.org/extensions/xep-0085.html
```
RemiBardon, line 15: Just for your information: the Swift markup language supports `- SeeAlso:`, but Xcode doesn't show it in Quick Help…
I think DocC supports it, but as I've not been able to compile the Prose documentation, I'm not sure.

Prose/ProseLib/Sources/ProseCoreTCA/Types/JID.swift
```diff
@@ -19,6 +19,12 @@ public extension JID {
     }
 }
 
+extension JID: Comparable {
+    public static func < (lhs: Self, rhs: Self) -> Bool {
+        lhs.rawValue < rhs.rawValue
+    }
+}
```
RemiBardon, line 26: Why do you need this? Is it for user-facing features? If it's the case, we should use a comparator like `String.Comparator(options: [.caseInsensitive, .diacriticInsensitive, .numeric]))` (changing the options as needed)

Prose/ProseLib/Sources/ProseCoreTCA/Types/Presence.swift
```diff
@@ -0,0 +1,99 @@
...
+        case probe
+        case error
+    }
+
+    enum Show: Equatable {
+        case away
+        case chat
+        case dnd
+        case xa
+    }
```
RemiBardon, line 34: Could you add a doc link to the XEPs?

RemiBardon: (Or add a link to the related `ProseCoreClientFFI.Xxx`)

Prose/ProseLib/Sources/TestHelpers/UUID+Incrementing.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+
+public extension UUID {
+    /// A deterministic, auto-incrementing "UUID" generator for testing.
+    static var incrementing: () -> UUID {
+        var uuid = 0
+        return {
+            defer { uuid += 1 }
+            return UUID(uuidString: "00000000-0000-0000-0000-\(String(format: "%012x", uuid))")!
```
RemiBardon, line 9: I'm not sure this is a valid UUIDv4, but `UUID` doesn't validate it IIRC, so it's OK.

Prose/ProseLib/Sources/TestHelpers/XCTestCase+Combine.swift
```diff
@@ -0,0 +1,91 @@
+import Combine
+import XCTest
+
+// Source: https://www.swiftbysundell.com/articles/unit-testing-combine-based-swift-code/
```
RemiBardon, line 4: Could you add this as `/// - Copyright:` on `await`? This way it appears in Quick Help.

Prose/ProseLib/Tests/ProseCoreTCATests/Helpers/BareJid+Tests.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+import ProseCoreClientFFI
+
+extension BareJid: ExpressibleByStringLiteral {
+    public init(stringLiteral value: String) {
+        self = try! ProseCoreClientFFI.parseJid(jid: value)
```
RemiBardon, line 6: `try!` is OK here, but (I think) a `fatalError` with a message would be better 💁🏻‍♂️

Prose/ProseLib/Tests/ProseCoreTCATests/ProseClientTests.swift
```diff
@@ -0,0 +1,357 @@
+import Combine
+import Foundation
+import ProseCore
+import ProseCoreTCA
+import TestHelpers
+import XCTest
+
+final class ProseClientTests: XCTestCase {
+    func testGroupsMessagesInChatsWhenReceiving() throws {
+        let date = Date.ymd(2_022, 6, 25)
```
RemiBardon, line 10: I should probably change the swiftformat rule to start at 5 digits

RemiBardon (approved):
Thank you very much for all this work!

Prose/ProseLib/Sources/App/AppReducer.swift
nesium: I know, although I find it less of a problem on macOS where you can still use the app. In Crisp there's a banner that you have to tap to trigger the request, so I'd expect that we would do something similar. I wouldn't view this as a final thing but something for our current use instead.

Prose/ProseLib/Sources/TestHelpers/XCTestCase+Combine.swift
nesium: And it's important that you see the copyright in quick help?

Prose/ProseLib/Sources/NotificationsClient/NotificationsClient+Live.swift
nesium: Delayed why?

RemiBardon: To reduce the app startup time?

nesium: Static vars, like lazy, are only evaluated when called. I also wouldn't think that `getNotificationSettings` is expensive as it even uses a callback to return the result.

PR #74, opened 2022-06-27
Update swiftformat settings

nesium:
We should run it, once all PRs are merged. But you can try it locally.

Can I persuade you to 2 spaces? 👻

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -1,4 +1,4 @@
---indent 4
+--indent 2
 --self insert
 --disable enumNamespaces
 --enable isEmpty,blockComments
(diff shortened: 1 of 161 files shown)


nesium:
I'm using 2 spaces in all of my projects, but we don't have to change, 4 spaces seems to be much more popular.

I think the default setting is fine for decimal grouping. As you've mentioned yourself 3 is drastic as it even formats 4 digit years.

RemiBardon:
That's why I was saying 3,5 instead of 3,4 (grouping by 3, starting from 5 digits)

nesium:
Ah, ok. Sure, if you prefer it that way.

PR #75, opened 2022-06-27
feat: Use `prose-core-views` release artifact (and bump version)

RemiBardon:
Now we're using [prose-core-views 0.2.3](https://github.com/prose-im/prose-core-views/releases/tag/0.2.3), and we don't need to clone the problect + build it, as we use release artifacts.

The new chat looks like:
Screenshot 2022-06-27 at 14 50 05


RemiBardon:
See https://prose-im.slack.com/archives/C03GGAZ69GT/p1656324035393179?thread_ts=1656173281.865569&cid=C03GGAZ69GT

valeriansaliou:
I can confirm what Marc said.

image

cURL would also nicely do it, and it comes in the standard macOS distribution, example:

```
curl -o archive.tar.gz 'http://server.tld/archive.tar.gz'
```

PR #77, opened 2022-07-07
feat: Add `EditProfileFeature` UI

RemiBardon:


RemiBardon:
The [Mac logo](https://developer.apple.com/licensing-trademarks/mac-logo/) present in the design artboards is subject to licensing, so I didn't add it. We will find another design.

The original design:
Screenshot 2022-07-07 at 10 58 35

The [Mac logo licensing notice](https://developer.apple.com/licensing-trademarks/mac-logo/):
Screenshot 2022-07-07 at 11 25 59

RemiBardon:
Dark mode looks perfect first try 👌🏼

RemiBardon:
Accessibility almost done ✅ I'm doing localization at the same time.

RemiBardon:
Most of the UI is:

`ContentSection`,
Screenshot 2022-07-08 at 12 11 55

`ThreeColumns`
Screenshot 2022-07-08 at 12 12 17

or `SecondaryRow`.
Screenshot 2022-07-08 at 12 12 44

This way, everything is accessible out of the box, and UI is coherent.

RemiBardon:
While doing this feature, I introduced a way to change the cursor. I'm not sure it's really worth it 🤔 If you feel like we don't need it and I should remove it, I get it.

RemiBardon:
I forgot to connect this view to the "Edit profile" button in the app. I'll do it before merge.

Prose/ProseLib/Sources/EditProfileFeature/Components/ThreeColumns.swift
```diff
@@ -0,0 +1,97 @@
...
+    self.label = label
+    self.column1 = { Text(verbatim: label) }
+    self.column2 = column2
+    self.column3 = column3
+  }
+
+  init(
+    _ label: String,
+    @ViewBuilder column2: @escaping () -> Column2
+  ) where Column1 == Text, Column3 == EmptyView {
```
nesium, line 40: Just tried this…

```swift
  init(
    _ label: String,
    @ViewBuilder column2: @escaping () -> Column2
  ) where Column1 == Text, Column3 == AnyView {
    self.label = label
    self.column1 = { Text(verbatim: label) }
    self.column2 = column2
    self.column3 = { AnyView(Color.clear.fixedSize()) }
  }
```

which gives the same result as using `ZStack`s below. Still not great, but feels less magic in my opinion.

nesium (approved):
Looks good. Thanks!

Indeed I vote for removing the `CursorClient`.

I've looked around in System Preferences and Mail and couldn't observe the same cursor behavior in these stock apps. It seems to be more a web thing (Slack does it). Also the delay feels weird if you pay notice until the cursor changes and it seems to not work correctly all of the time. Hovering over the items you can get in a state where the hand cursor isn't visible even if it should.

Regarding the `ThreeColumns` view and every other custom container view, I propose that we don't store escaping closures but resolve these in the initializer instead, for the reasons outlined [here](https://rensbr.eu/blog/swiftui-escaping-closures/).

nesium:
One more thing. Could you please add all your preview apps to the `build_preview_apps` Make task? … and run `make preflight` before pushing. Otherwise these preview apps have a tendency to break over time which is something we should prevent.

RemiBardon:
> Indeed I vote for removing the `CursorClient`.

I have the same arguents, so I will remove it.

> I propose that we don't store escaping closures but resolve these in the initializer instead

Oh yes, I had never thought of it that way, I always wondered why Apple's closures were not `@escaping`

Prose/ProseLib/Sources/EditProfileFeature/EditProfileScreen.swift
```diff
@@ -0,0 +1,210 @@
...
+    switch action {
+    case .saveProfileTapped:
+      logger.trace("Save profile tapped")
+      return .none
+
+    case .cancelTapped:
+      logger.trace("Cancel profile editing tapped")
+      return .none
+
+    default:
```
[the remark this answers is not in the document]
RemiBardon: I don't really like it either, I thought I had seen you do it and I thought you would prefer it that way. Will revert this (and do it in #80 too, where I did the same)

Prose/ProseLib/Sources/EditProfileFeature/Components/ThreeColumns.swift
RemiBardon: Indeed, if feels less magic, but I don't know if there's a downside to using `Color.clear` (rendering, hit testing…).

`AnyView` seems ok in this case, as it's not some content-related view.

Prose/ProseLib/Sources/EditProfileFeature/AuthenticationView.swift
```diff
@@ -0,0 +1,161 @@
...
+              ThreeColumns(l10n.MfaToken.Header.label) {
+                Button { viewStore.send(.disableMFATapped) } label: {
+                  Text(verbatim: l10n.MfaToken.DisableMfaAction.label)
+                    .frame(maxWidth: .infinity)
+                }
+                .controlSize(.large)
+              }
+              SecondaryRow(l10n.MfaStatus.Header.label) {
+                HStack(spacing: 4) {
+                  // TODO: Change this to a more general purpose indicator
```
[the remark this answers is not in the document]
RemiBardon: I mean, we have indicator dots all around the UI. This one has no specific meaning, so it could be refactored out so we can use it in other places.

I already created `ConnectionStatusIndicator` and `OnlineStatusIndicator`, maybe one could be renamed or a third one could be created.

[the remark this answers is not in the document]
RemiBardon: `ConnectionStatus` is `(disconnected|connected)`, but maybe we could rename it and change this to `(on|off)` or something.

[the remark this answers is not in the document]
nesium: I don't know. The TODO just isn't very specific and I'm generally not a big fan to treat code like a ticketing system.

Prose/ProseLib/Sources/EditProfileFeature/Components/ThreeColumns.swift
RemiBardon: I just rememebered the reason why I did this:
One problem with this is that only this initializer now handles the problem. If you manually call `init(label:column1:column2:column3:)` giving an `EmptyView` (which was my case before the refactor), it breaks the layout.

nesium: Isn't that true for every other SwiftUI container as well? Garbage in, garbage out.

RemiBardon:
```swift
VStack(alignment: .leading) {
  Label {
    Text("test")
  } icon: {
    Text("icon")
  }
  Label {
    Text("test")
  } icon: {
    EmptyView()
  }
  Label {
    EmptyView()
  } icon: {
    Text("icon")
  }
}
.padding()
```
Screenshot 2022-07-12 at 11 32 29

Same in a `List`, looks like you're right. I'll use your proposition.

Prose/ProseLib/Sources/EditProfileFeature/EditProfileScreen.swift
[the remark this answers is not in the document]
RemiBardon: Ah… I will do it in batch for all reducers in another PR then, I noted it.

RemiBardon:
@nesium All good?

nesium:
LGTM!

RemiBardon:
Oh no, still need to integrate it in the app! Lucky I added a task in the PR description 😬

RemiBardon:
@nesium WDYT? I didn't use a `ViewState` in `SidebarView`, tell me if I need to do it.

Prose/ProseLib/Sources/SidebarFeature/Footer/Footer.swift
```diff
@@ -28,7 +39,7 @@ struct Footer: View {
         // User avatar
         FooterAvatar(store: self.store.scope(state: \.avatar, action: Action.avatar))
 
-        WithViewStore(self.store, removeDuplicates: ==) { viewStore in
+        WithViewStore(self.store) { viewStore in
```
nesium, line 42: Why not pull these properties below into the `ViewState` as well?

Prose/ProseLib/Sources/SidebarFeature/Footer/Footer.swift
```diff
@@ -61,23 +93,49 @@ struct Footer: View {
...
-    environment: { $0 }
+    environment: { _ in () }
   ),
   footerAvatarReducer.pullback(
-    state: \.avatar,
+    state: \FooterState.avatar,
     action: CasePath(FooterAction.avatar),
+    environment: { _ in () }
+  ),
+  editProfileReducer.pullback(
```
nesium, line 108: Would you mind referencing the ticket `https://github.com/pointfreeco/swift-case-paths/issues/86` here. Hopefully we can remove that workaround in the future know why it exists in the meanwhile.

Prose/ProseLib/Sources/SidebarFeature/Footer/Footer.swift
```diff
@@ -86,28 +144,49 @@ struct FooterState: Equatable {
...
     self.statusMessage = statusMessage
     self.avatar = avatar
     self.actionButton = actionButton
+    self.sheet = sheet
+  }
+}
+
+public extension FooterState {
+  enum Sheet: Equatable {
+    case editProfile(EditProfileState)
```
nesium, line 170: Are we expecting to show more sheets here or could this be a regular property?

Prose/ProseLib/Sources/SidebarFeature/Footer/Footer.swift
```diff
@@ -86,28 +144,49 @@ struct FooterState: Equatable {
...
+
   init(
     credentials: UserCredentials,
     teamName: String = "Crisp",
     statusIcon: Character = "🚀",
     statusMessage: String = "Building new features.",
     avatar: FooterAvatarState = .init(avatar: .placeholder),
-    actionButton: FooterActionMenuState = .init()
+    actionButton: FooterActionMenuState = .init(),
+    sheet: Sheet? = nil
```
nesium, line 156: I'm not a big fan of these extensive initializers. Do we really need to inject the sheet property? Generally I think the initializers should only accept what they really need. Don't allow any misuse in production code. Everything else can be set via properties if we need to do so in tests. And these properties should be `internal` at most (which is already the case here).

Prose/ProseLib/Sources/ProseUI/BooleanIndicator.swift
```diff
@@ -0,0 +1,77 @@
+//
+// This file is part of prose-app-macos.
+// Copyright (c) 2022 Prose Foundation
+//
+
+import Assets
+import ProseCoreTCA
+import SwiftUI
+
+public struct BooleanIndicator: View {
```
nesium, line 10: That's _really_ generic. I have no idea what a `BooleanIndicator` would look like. Also what's next? A `IntIndicator` or a `StringIndicator`? I'm a bit snarky here to press my point. Maybe the name could more reflect what the view draws instead of the underlying value. Like a LED light or something? Otherwise you could indeed call a label a `StringIndicator`.

nesium (commented):
What I would love to have in terms of keyboard navigation is:
- Move up and down in the sidebar if it has the immediate focus (like the sidebar in the main window)
- Press ESC to dismiss the sheet

Is there an easy way to handle that? E.g. use a List for the sidebar.

RemiBardon:
@nesium I'll fix all of this some time today.

Just quickly:

> What I would love to have in terms of keyboard navigation is:
> 
> Move up and down in the sidebar if it has the immediate focus (like the sidebar in the main window)

That's what I wanted too…

> use a List for the sidebar.

I've tried hard, it's impossible. See https://prose-im.slack.com/archives/C03G1PS24B1/p1657127700054459?thread_ts=1657118158.691269&cid=C03G1PS24B1

> - Press ESC to dismiss the sheet
> 
> Is there an easy way to handle that?

This one I don't know how easy/hard it is using SwiftUI… I bet it's not very easy, but we might be able to do it anyway.

Prose/ProseLib/Sources/SidebarFeature/Footer/Footer.swift
RemiBardon: I guess I won't need to 🙃

RemiBardon: I'll update the dependency instead

Prose/ProseLib/Sources/SidebarFeature/Footer/Footer.swift
RemiBardon: For no reason, just forgot as it was not on my "bad performance" path. (And for this subview size, we don't usually create a `ViewState`)

Prose/ProseLib/Sources/ProseUI/BooleanIndicator.swift
RemiBardon: You're right, it's too generic. I had no other name in mind. `LEDIndicator` is a good one 🙂

Prose/ProseLib/Sources/EditProfileFeature/Sidebar/Sidebar.swift
```diff
@@ -0,0 +1,149 @@
...
+  Void
+>.combine([
+  sidebarHeaderReducer.pullback(
+    state: \SidebarState.header,
+    action: CasePath(SidebarAction.header),
+    environment: { $0 }
+  ),
+  Reducer { state, action, _ in
+    if case .row(_, .rowTapped) = action {
+      state.rows[id: state.selection]?.isSelected = false
```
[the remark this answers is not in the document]
RemiBardon: `state.selection` is updated when we tap a row… but I'm so used to `selection`s being `BindableState`s that I didn't think this time I could just put it before `state.selection = id` 🙄

Thank you for this catch!

Prose/ProseLib/Sources/SidebarFeature/Footer/Footer.swift
RemiBardon: Yes, from "Account settings"

Screenshot 2022-07-13 at 18 17 51

RemiBardon:
I'll rebase quickly, and then address the fact that the sidebar is not a `List`.

As discussed on Slack, I made a mistake by interpreting the blue in the artboards for a "design blue", while it was just the default macOS accent color.

Therefore, I can remove the highlighting part (not usual on macOS anyway), and use a regular `List`. I like it.

I'm going to fix all the blue in another PR, but in this one I'll fix the sidebar at least. It will have a different color from some of the buttons, but I'll change this very soon!

RemiBardon:
There are a few issues I see, but they're minor and I suppose we can forget them:

When the window isn't first responder, the white on light gray lacks some contrast
Screenshot 2022-07-13 at 19 35 14

The sheet can be resized to the point where the sidebar is hidden, but this is reversible so it's not a real problem
Screenshot 2022-07-13 at 19 27 03

I probably put some `maxWidth` here at some point 🤔
Screenshot 2022-07-13 at 19 26 57

And the selection binding not being updated before `touchUpInside`, as shown in the video before

RemiBardon:
> I probably put some maxWidth here at some point 🤔

No, I didn't set `alignment: .leading` on the header. Let me fix this quickly.

RemiBardon:
The last commit was the most important! When changing the routing, I forgot to reconnect the child `Reducer`s!

RemiBardon:
@nesium Everything's good now?

RemiBardon:
(`make preflight` succeeds)

nesium:
Let's limit the width of the EditProfile sheet, so that the list doesn't collapse. These values seem to do the trick…

```swift
.frame(minWidth: 900, maxWidth: 1500)
```

RemiBardon:
I added `.frame(minWidth: Self.minContentWidth + Sidebar.minWidth + 1)` (which adds up to 809 FYI).

(The `+ 1` is for the separator, it would still collapse because of it otherwise 😅)

PR #78, opened 2022-07-11
feat: Add a way to display search suggestions

RemiBardon:
It was initially in #76, but I separated the two features.


RemiBardon:
> Selects the next responder when Tab is pressed

@nesium Why wouldn't this already work? We do not intercept <kbd>Tab</kbd> events 🤔

RemiBardon:
There are too many commits cluttering the PR, let me rebase this correctly.

RemiBardon:
This is nice 🙂

https://user-images.githubusercontent.com/37386490/181022933-f0fedc22-365a-4613-9df3-8ae3ea270622.mov

No more height computation, no more manual clipping, no more background color…

PR #82, opened 2022-07-12
Fix chat initialization

nesium:
This fixes the bug where messages where not displayed when the chat was shown for the first time. UI Testcase coming right up in the next PR (#83).


RemiBardon (approved):
Thank you for this catch 🙂

PR #83, opened 2022-07-12
UI test setup

nesium:
This PR adds a UI test setup.

Right now you'll see a failing test if you run the UI tests (fix is in #82). You can do so by either selecting the Prose target and selecting either the `AllTests` or `UITests` test plan (see below), by selecting the UITestHost target and running from there or by running `make test`.

Screenshot 2022-07-12 at 17 11 10

The setup might be a bit unconventional since the (current batch of) UI tests don't run against the app target. The reason is that running against the app target makes it very difficult to inject test data without running a full-blown XMPP server next to it. Instead I've added the (aforementioned) UITestHost target which can be parameterized with an environment variable to run different test scenarios.

You can see an example test scenario in `ProseLib/TestHostApp/TestCases/RosterSelection.swift`. There's a pretty basic mock environment set up just to display some roster items and display a message in a chat. The actual test case is in `RosterSelectionTests/RosterSelectionTests.swift` where the application is run by specifying the scenario/test case with…

```swift
let app = XCUIApplication.launching(testCase: "roster-selection")
```

The actual environment variable -> Scene matching is done in `ProseLib/TestHostApp/Helpers/TestScene.swift` with a switch…

```swift
…
switch ProcessInfo.processInfo.environment["test-case"] {
case "roster-selection":
  RosterSelection()
…
```

I bet there's room for improvement to not have a stringly typed API, but for now you can fairly easy set up another test case.

You can also run test cases standalone from the UITestHost target by specifiying the `test-case` environment variable in the scheme.

I've also added a `Mocks` (package) target that can not only be used in tests but also in preview apps. In there you'll find a JSON with a bunch of random users from IIRC randomuser.me (I had that around for a while) in case you need some random users.


nesium:
Let me merge the other PRs and rebase this real quick.

nesium:
The test now succeeds

Prose/ProseLib/Sources/Mocks/Message.swift
```diff
@@ -0,0 +1,29 @@
...
+// Copyright (c) 2022 Prose Foundation
+//
+
+import Foundation
+import ProseCoreTCA
+
+public extension Message {
+  static func mock(
+    from: JID = "jane.doe@prose.org",
+    id: String = "message-id",
```
RemiBardon, line 12: Does it make sense to have a static default id? Won't it cause conflicts?
I usually do `UUID().uuidString`, keeping only a subset if needed (for length).

Prose/ProseLib/Sources/Mocks/Publisher.swift
```diff
@@ -0,0 +1,15 @@
+//
```
RemiBardon, line 1: Could the file be named something like `Publisher+Ext.swift` instead?

RemiBardon: Or `+Effect`, or `+TCA`

Prose/ProseLib/Sources/TestHostApp/Helpers/TestScene.swift
```diff
@@ -0,0 +1,70 @@
...
+          Missing or unknown test case.
+
+          If you're trying to run a UI Test, make sure to specify the desired testcase in \
+          your XCUIApplication with:
+
+          app.launchEnvironment = ["test-case": "roster-selection"]
+
+          If you're trying to run the app manually, add a Environment Variable "test-case" to \
+          the current scheme.
+          """)
```
RemiBardon, line 34: Smart way of showing the error without aborting 💡

Prose/ProseLib/Sources/TestHostApp/Helpers/TestScene.swift
```diff
@@ -0,0 +1,70 @@
...
+    SwizzleImplementations(
+      in: NSScrollView.self,
+      originalSelector: #selector(flashScrollers),
+      swizzledSelector: #selector(prose_flashScrollers)
+    )
+  }
+
+  @objc func prose_flashScrollers() {
+    // Do nothing…
+  }
```
RemiBardon, line 69: What is this? Will this make something one day?

RemiBardon: Oh no, I guess you used it to swizzle a method

RemiBardon (approved):

RemiBardon:
Thank you for this 🙂 As you said, one day we could remove the stringy API, but we'll do it when it bothers us too much, and we still need to write tests for it to happen 🙈

Prose/ProseLib/Sources/TestHostApp/Helpers/TestScene.swift
nesium: Correct. That's the swizzled method.

PR #85, opened 2022-07-26
feat: Create `CustomList`

RemiBardon:
Following a Slack discussion on https://github.com/prose-im/prose-app-macos/pull/78#issuecomment-1190102480, we decided to reimplement a basic SwiftUI `List` using a `VStack`.

I managed to recreate a good portion of the `List` API, or at least what we will need of it.

Here are the results:

Screenshot 2022-07-26 at 10 40 03

When the window is not key:

Screenshot 2022-07-26 at 10 40 12

With sections (supports `.headerProminence(.increased)`):

Screenshot 2022-07-26 at 10 41 02

I had to create a custom selection shape, to get the same effect as `List` when multiple items are selected and touch each other:

Screenshot 2022-07-26 at 10 44 04

There is no support for keyboard navigation. This must be implemented higher in the hierarchy, or I can implement it here as well if needed. The higher implementation will be necessary anyway, when the list is not first responder, which is the only place where we will use `CustomList`, so I don't think we need to duplicate this code.


RemiBardon:
I'll just add an initializer which takes sectionned content but a single selection, I need it for #78.

RemiBardon:
I just tested with the attached window, and it works (using `.fixedSize(horizontal: true, vertical: false)` for rows and header. I did not enforce this in my `CustomList` implementation, should I?

Also, the attached window is not key, so it gives this bad result. However, this must be fixed in the other PR, and https://github.com/smic/SuggestionsDemo seems to handle it well. We'll see.

Screenshot 2022-07-26 at 15 08 56
Screenshot 2022-07-26 at 15 10 36
Screenshot 2022-07-26 at 15 10 50

RemiBardon:
The magic of SwiftUI: I just added `.environment(\.controlActiveState, .active)` to the attached window, and it works 🥳

Screenshot 2022-07-26 at 15 40 50

@nesium Should I add this modifier on `AttachedWindow` itself or are there cases where we won't want to enforce this?

PR #92, opened 2022-08-01
feat: Implement message bar emoji picker

RemiBardon:
As we don't have a reference to any `NSView`, I have to use `NSApp.orderFrontCharacterPalette(nil)`. If no text field is focused, the character palette will be useless, so I chose to disable the emoji button if the message bar text field is not focused:
Screenshot 2022-08-01 at 11 55 02
When it's focused, the button shows again:
Screenshot 2022-08-01 at 11 55 10
When tapped, the button shows the palette, at the correct location:
Screenshot 2022-08-01 at 11 55 44
Screenshot 2022-08-01 at 11 56 05

And finally, the button can be used multiple times without ending in a weird state.


valeriansaliou:
LGTM to me in terms of UX.

RemiBardon:
See https://github.com/prose-im/prose-app-macos/pull/91#issuecomment-1206548535

PR #95, opened 2022-08-05
chore: Add github workflow

nesium:


.github/workflows/ios.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: make lint
+
+      - name: Run unit tests
+        run: make test
+
+      - name: Build preview apps
+        run: make build_preview_apps
+
+      - name: Build release scheme
+        run: release_build
```
RemiBardon, line 38: Shouldn't this be `make release_build`?

RemiBardon (changes requested):

.github/workflows/ios.yml
nesium: Thanks. I didn't request a review even though you're right. But as you can see the workflow isn't even running yet.

RemiBardon: No problem, thanks for the PR, I was happy to see this 🙂

RemiBardon (approved):

PR #98, opened 2022-08-08
feat: Bump `prose-core-views` to `0.10.0`

RemiBardon:
This PR bumps core-views to [`0.10.0`](https://github.com/prose-im/prose-core-views/releases/tag/0.10.0).

Now message reactions are correctly ordered.


Prose/ProseLib/Sources/ProseCoreTCA/OrderedCollections+Ext.swift
```diff
@@ -0,0 +1,43 @@
+//
+// This file is part of prose-app-macos.
+// Copyright (c) 2022 Prose Foundation
+//
+
+import OrderedCollections
+
+public extension Dictionary {
+  @discardableResult mutating func prose_toggle<Element>(
```
nesium, line 9: We have now 5 public, untested, (4) undocumented `prose_toggle` extensions and one caller. To me that begs the question if we shouldn't just inline the implementation in that one place where it is used.

Prose/ProseLib/Sources/ProseCoreTCA/Types/Message.swift
```diff
@@ -93,14 +94,16 @@ extension MessageKind {
 }
 
 public struct MessageReactions: Equatable {
-  var reactions: [Reaction: Set<JID>]
+  public typealias WrappedValue = OrderedDictionary<Reaction, OrderedSet<JID>>
```
nesium, line 97: Why do the JIDs need to be ordered as well?

Prose/ProseLib/Sources/ConversationFeature/Chat/Chat.swift
```diff
@@ -42,7 +47,13 @@ struct ProseCoreViewsMessage: Equatable, Encodable {
       jid: message.from.jidString,
       name: message.from.jidString
     )
-    self.reactions = message.reactions
+    self.reactions = message.reactions.reactions.map(Reaction.init(from:))
```
nesium, line 50: Couldn't we adjust the `encode(to:)` method of `MessageReactions` or alternatively get rid of it?

nesium (approved):

Prose/ProseLib/Sources/ConversationFeature/Chat/Chat.swift
RemiBardon: I wanted to keep a layer with types from prose-core-views. Nothing says `MessageReactions` is encoded in JSON as the web view needs it. And we shouldn't have to change our internal model if the web view API changes. I agree, right now we have the same structure, but nothing says it will stay so.

Prose/ProseLib/Sources/ProseCoreTCA/Types/Message.swift
RemiBardon: You're right, there are no reason right now because we're not displaying the names, but as it's chronologically ordered I thought it'd make sense to keep the order.

Prose/ProseLib/Sources/ProseCoreTCA/OrderedCollections+Ext.swift
RemiBardon: I don't know. It could almost be generic, but I found no way not to duplicate the code.
Maybe I can remove some non-ordered extensions at least, which we do not use.

Prose/ProseLib/Sources/ProseCoreTCA/Types/Message.swift
RemiBardon: So should I use `Set` or should I add a doc comment?

Prose/ProseLib/Sources/ConversationFeature/Chat/Chat.swift
nesium: By manually implementing the `encode(to:)` method we already have a layer of abstraction.

You can leave it like that, but the thing is that I'd like to have our models as close as possible to the (core-)view models. Over time - with conversations growing - mapping over each message (and now reactions) gets quite expensive since you have to allocate a new array each time and due to the reactive nature of our architecture this happens for every small change to the messages.

I've noticed that this becomes especially apparent when you're scrolling and inserting loaded older messages. The allocations can then be noticeable when the scrollview drops frames.

The JSON marshalling probably has the most overhead right now, I just want to give you my rationale.

nesium: If you leave it like that please remove the `MessageReactions` conformance to `Encodable` though since its not used anymore.

Prose/ProseLib/Sources/ProseCoreTCA/Types/Message.swift
nesium: Please make it as simple as possible and use a Set.

Prose/ProseLib/Sources/ConversationFeature/Chat/Chat.swift
RemiBardon: I get your point, and you're right we could change this.

My issue with it is just that `encode(to:)` is not linked by any mean to JSON. Only `JSONEncoder` and `JSONSerialization` are. Creating a custom `encode(to:)` just for the web view means we will never be able to encode for anything else (unless it uses the exact same format, which is highly improbable). If we could have some other function instead, which specifically targets the web view, I'm all for it!

Do you have an idea? What's better than mapping it to another struct? Using `[String: Any]` and `JSONSerialization`?

nesium: The shape of the struct is not really the problem. The two problems that I'm seeing are:
1. `messages.map` is called for each change to `messages` and `Collection.map` allocates a new array for the mapped results.
2. JSON encoding is expensive but I wouldn't want to address that now. We can use proto- or flatbuffers at some point.

Overall I would like to treat the models from the ProseCoreTCA targets as our view models. And a requirement of our view models is that they need to be able to be serialized to JSON.

Sure if we'd want to store these for caching purposes or something and would want to have a different format there or make sure that it is stable we'd need to split the two serialization paths (for storage and for the webview).

This would most likely be done one layer deeper in the core lib though where we have separate models.

But that's getting too complicated for that format here already. I'll make a note so that we can address that briefly in our call.

RemiBardon: Should I open a new issue and fix this struct duplication tomorrow (or sometime soon) or should I handle it in this PR?

nesium: "Sometime soon" sounds good to me :)

RemiBardon: I created #103. I will resolve this conversation and merge the PR now.

PR #100, opened 2022-08-08
feat: Bump `prose-core-views` to `0.10.2`

RemiBardon:
This PR bumps core-views to [`0.10.2`](https://github.com/prose-im/prose-core-views/releases/tag/0.10.2).

Now clicks on the web view sometimes send a frame along the click point.

This means the reactions popup is now placed by the system around the reactions button instead of being placed where the user clicked:

- Default
  Screenshot 2022-08-08 at 18 06 04
- No space on top
  Screenshot 2022-08-08 at 18 10 11
- No space on the right and top
  Screenshot 2022-08-08 at 18 06 15


Prose/ProseLib/Sources/ConversationFeature/Chat/ChatReducer.swift
```diff
@@ -26,19 +26,30 @@ struct ChatState: Equatable {
...
+  let x, y: Double
+  var cgPoint: CGPoint { CGPoint(x: self.x, y: self.y) }
+}
 
+struct CoreViewsFrame: Equatable, Decodable {
+  let x, y, width, height: Double
+  var cgRect: CGRect { CGRect(x: self.x, y: self.y, width: self.width, height: self.height) }
+}
+
+struct CoreViewsOrigin: Equatable, Decodable {
```
nesium, line 44: `Origin` is a bit ambiguous here. Maybe we should follow suit and also call it `EventOrigin` like in the core views repo?

nesium (approved):

Prose/ProseLib/Sources/ConversationFeature/Chat/ChatReducer.swift
RemiBardon: Ok, it's just that I thought we wanted to create CoreViews- mapping structs. `EventOrigin` is quite the opposite, it reads very general. What about `CoreViewsEventOrigin`?

nesium: Yes, that's what I meant.

PR #101, opened 2022-08-08
fix: Fix memory leak of ChatView.Coordinator and ChatWebView

nesium:
When you switched back and forth between e.g. "Direct Messages" and any chat in the sidebar for each selection of a chat a   ChatView.Coordinator and WKWebView was leaked.

### DIFF
--- a/Prose/ProseLib/Sources/ConversationFeature/Chat/Chat.swift
+++ b/Prose/ProseLib/Sources/ConversationFeature/Chat/Chat.swift
@@ -199,11 +199,14 @@ struct ChatView: NSViewRepresentable {
       // but still allow starting with a non-nil value (which `dropFirst()` would prevent).
       .drop(while: { $0 == nil })
       .removeDuplicates()
-      .sink { menuState in
+      .sink { [weak coordinator = context.coordinator] menuState in
+        guard let coordinator = coordinator else {
+          return
+        }
         if let menuState: MessageMenuState = menuState {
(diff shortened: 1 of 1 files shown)


Prose/ProseLib/Sources/ConversationFeature/Chat/Chat.swift
```diff
@@ -199,11 +199,14 @@ struct ChatView: NSViewRepresentable {
...
       .drop(while: { $0 == nil })
       .removeDuplicates()
-      .sink { menuState in
+      .sink { [weak coordinator = context.coordinator] menuState in
+        guard let coordinator = coordinator else {
+          return
+        }
         if let menuState: MessageMenuState = menuState {
-          self.showMenu(menuState, on: webView, context: context)
+          self.showMenu(menuState, on: webView, coordinator: coordinator)
```
RemiBardon, line 207: Isn't it a problem that `webView` is sent here? Shouldn't we add `weak webView` too?

nesium: No ;)

RemiBardon: The answer is no: The coordinator was captured, which prevented the `AnyCancellable`s from being canceled. Now they are, which frees the web view.

RemiBardon (approved):
Thank you for this catch!

PR #102, opened 2022-08-09
feat: Bump `prose-core-views` to `0.11.0`

RemiBardon:
This PR bumps core-views to [`0.11.0`](https://github.com/prose-im/prose-core-views/releases/tag/0.11.0).

Now the message actions in the top-right corner of a message are always visible when the reactions picker is open (same for the menu, but because it blocks the main thread).

This PR also introduces diffing to avoid reloading the whole chat when updating a message (it was now necessary).


Prose/ProseLib/Sources/ConversationFeature/Chat/Chat.swift
```diff
@@ -254,20 +262,64 @@ struct ChatView: NSViewRepresentable {
...
+    let ids = OrderedSet<Message.ID>(messages.map(\.id))
+    let oldIds: OrderedSet<Message.ID> = coordinator.messages.ids
+
+    let insertedIds: OrderedSet<Message.ID> = ids.subtracting(oldIds)
+    let removedIds: OrderedSet<Message.ID> = oldIds.subtracting(ids)
+    let updatedIds: [Message.ID] = ids.intersection(oldIds)
+      .filter { messages[id: $0] != oldMessages[id: $0] }
+
+    for messageId in removedIds {
+      self.rectractMessage(withId: messageId, in: webView)
```
nesium, line 285:
```suggestion
      self.retractMessage(withId: messageId, in: webView)
```

Prose/ProseLib/Sources/ConversationFeature/Chat/Chat.swift
```diff
@@ -254,20 +262,64 @@ struct ChatView: NSViewRepresentable {
...
+  func updateMessage(to message: ProseCoreViewsMessage, in webView: WKWebView) {
+    logger.trace("Updating 1 message…")
+
+    let script = """
+    MessagingStore.update(\(Self.unsafeJson(message.id)), \(Self.unsafeJson(message)));
+    """
+    webView.evaluateJavaScript(script, domain: "Update message")
+  }
+
+  func rectractMessage(withId messageId: ProseCoreViewsMessage.ID, in webView: WKWebView) {
```
nesium, line 316:
```suggestion
  func retractMessage(withId messageId: ProseCoreViewsMessage.ID, in webView: WKWebView) {
```

Prose/ProseLib/Sources/ConversationFeature/Chat/Chat.swift
```diff
@@ -372,6 +442,23 @@ struct ChatView: NSViewRepresentable {
...
+      return "[]"
+    }
+    return String(data: jsonData, encoding: .utf8) ?? "[]"
+  }
+
+  /// - Note: This method is named unsafe because it uses force unwraps (`!`).
+  ///         The object is escaped properly.
+  static func unsafeJson<T: Encodable>(_ object: T) -> String {
+    let jsonData: Data = try! JSONEncoder().encode(object)
+    return String(data: jsonData, encoding: .utf8)!
```
nesium, line 460: We could at least crash with `fatalError` and a message that prints the type that was tried to be encoded so that we know what went wrong more easily.

Prose/ProseLib/Sources/ConversationFeature/Chat/Chat.swift
```diff
@@ -254,20 +262,64 @@ struct ChatView: NSViewRepresentable {
...
-    let json = String(data: jsonData, encoding: .utf8) ?? "[]"
+    let messages = IdentifiedArrayOf<ProseCoreViewsMessage>(uniqueElements: messages)
+    let oldMessages: IdentifiedArrayOf<ProseCoreViewsMessage> = coordinator.messages
+    defer { coordinator.messages = messages }
+
+    let ids = OrderedSet<Message.ID>(messages.map(\.id))
+    let oldIds: OrderedSet<Message.ID> = coordinator.messages.ids
+
+    let insertedIds: OrderedSet<Message.ID> = ids.subtracting(oldIds)
+    let removedIds: OrderedSet<Message.ID> = oldIds.subtracting(ids)
```
nesium, line 280: Please extract that diffing logic and add a few rudimentary tests at least. Doesn't need to be generic or anything and can live in the (yet to be introduced) ConversationFeatureTests.

nesium (commented):
The interaction between the prose-core-views and our code grows ever more untestable unfortunately. 

I think we should add a Swift wrapper for the `MessageStore` which has the same methods but accepts Swift objects and converts the method call into a Javascript string internally.

The `evaluateJavaScript` method would be injected into the `MessageStore` wrapper. That way we could at least run snapshot tests on the generated strings.

Prose/ProseLib/Sources/ConversationFeature/Chat/Chat.swift
RemiBardon: For some reason, GitHub doesn't show me this one in the file explorer. I can't commit the other proposition at the same time from here. I will do this in a commit on my machine (it won't set the co-author).

RemiBardon: Thank you for the catch by the way!

Prose/ProseLib/Sources/ConversationFeature/Chat/Chat.swift
RemiBardon: I made it generic to make testing easier

RemiBardon:
I was about to ask you a review again, I forgot your review message 🙄 I always try to resolve conversations, and sometimes I forget the message

RemiBardon:
It's fixed now. @nesium WDYT?

nesium (approved):

PR #106, opened 2022-08-11
chore: Avoid useless mapping to `ProseCoreViews.Message`

### LINKED ISSUE #103: Avoid useless mapping of models from ProseCoreTCA to models from prose-core-views
See https://github.com/prose-im/prose-app-macos/pull/98#discussion_r941436046.

Mapping models causes a lot of useless overhead. ProseCoreTCA models could be `Encodable` using the format specified by `prose-core-views`.

RemiBardon:
Closes #103

@nesium Is there anything else I should remove?


Prose/ProseLib/Tests/ProseCoreTCATests/TypesEncodingTests.swift
```diff
@@ -0,0 +1,106 @@
...
+    XCTAssertTrue(expectedOneOf.contains(string))
+  }
+
+  func testJIDEncoding() throws {
+    func test(_ jid: JID, equals expected: String) throws {
+      try self.check(jid, equals: expected)
+    }
+
+    try test("test@prose.org", equals: #""test@prose.org""#)
+    try test("test@prose.org/resource", equals: #""test@prose.org""#)
```
nesium, line 35: Try to keep the `XCTAssert…` as close as possible to the actual test. If it's not possible at least pass #file and #line through to it. Otherwise test failures are not shown in the right position - they are, just where you wouldn't expect them.

In this case you could return the JSON string from your `check` method and do the XCTAssertEqual inline without really losing anything.

Prose/ProseLib/Tests/ProseCoreTCATests/TypesEncodingTests.swift
```diff
@@ -0,0 +1,106 @@
...
+    try test(
+      ["🧪": [], "😀": ["tests1@prose.org"]],
+      equals: #"[{"authors":["tests1@prose.org"],"reaction":"😀"}]"#
+    )
+    // Reaction authors is stored in a `Set`, so we can't predict the output
+    try test(
+      ["2️⃣": ["tests1@prose.org", "tests2@prose.org"]],
+      in: [
+        #"[{"authors":["tests1@prose.org","tests2@prose.org"],"reaction":"2️⃣"}]"#,
+        #"[{"authors":["tests2@prose.org","tests1@prose.org"],"reaction":"2️⃣"}]"#,
```
nesium, line 72: Interesting approach. We could also sort the authors when encoding. That would probably have a negligible performance impact since it only affects reactions with multiple authors - although that might depend on the size of the workspace.

Another approach would be to pass the view only what it needs, which is the emoji, a count and a flag if one of those reactions is mine (current user).

nesium (approved):
Great. Thanks!

Prose/ProseLib/Tests/ProseCoreTCATests/TypesEncodingTests.swift
RemiBardon: Or, as I was doing a few days ago, use an `OrderedSet` 🙃 For the same reason as in my next sentence, We will go back to an `OrderedSet` one day or another 😌

> Another approach would be to pass the view only what it needs

At some point we will show this data in the view, so it's better like this

RemiBardon: > We could also sort the authors when encoding

I thought about it, but it's only useful for the tests, and you didn't want me to use `OrderedSet`, so I believed it wasn't a good idea

Prose/ProseLib/Tests/ProseCoreTCATests/TypesEncodingTests.swift
RemiBardon: I will do that

Prose/ProseLib/Tests/ProseCoreTCATests/TypesEncodingTests.swift
nesium: Ok, I put that here for the record. I completely forgot about the feature where you could see the author of each reaction when hovering over the reaction tag. And after a quick test with you on Slack these seem to be ordered by timestamp. So you were right the first time to use a OrderedSet and we should reintroduce that.

Issue #46, opened 2022-06-17
Messaging: Base messaging


valeriansaliou:
* [x] Integrate the messaging Web view from the [prose-core-views](https://github.com/prose-im/prose-core-views)
  * Description: The messaging Web view is ready, include it w/ static data

* [x] Pass store data to the messaging Web view
  * Description: Pass store data from Swift to the JS runtime. We should find a way to spawn a single WebKit background process and re-use the same view again and again when the user switches conversation, by simply updating the VueJS store in the view (this avoids re-parsing the JS everytime).

* [x] Ability to send simple text messages and receive them from the network
  * Description: Send messages w/ the `chat` type over XMPP, and receive `chat` messages back.
  * XEPs:
    * RFC3921: Exchanging Messages — https://xmpp.org/rfcs/rfc3921.html#messaging
    * XEP-0308: Last Message Correction — https://xmpp.org/extensions/xep-0308.html
    * XEP-0085: Chat State Notifications — https://xmpp.org/extensions/xep-0085.html
    * XEP-0280: Message Carbons — https://xmpp.org/extensions/xep-0280.html

* [x] Fetch messaging history from the server using MAM (if the server supports MAM) and populate the store
  * Description: Do this upon opening the conversation for the first time since connection broke.
  * XEPs:
    * XEP-0313: Message Archive Management — https://xmpp.org/extensions/xep-0313.html
    * XEP-0441: Message Archive Management Preferences — https://xmpp.org/extensions/xep-0441.html
  
* [x] Support message reactions
  * Description: The WebView should contain a button (on hover?) to add a reaction. Clicking the button would show a native menu. We'd start with UTF emojis only (no user-customized). Maybe a set of 10-20 something emojis would suffice as a start before with go all in with an emoji picker that has multiple pages and a search bar.
  * XEPs:
    * XEP-0444: Message reactions — https://xmpp.org/extensions/xep-0444.html

nesium:
I haven't done anything about Message Carbons yet. My understanding is that the client can just enable/disable this and the server manages the distribution. But I'll read that XEP once again and add a method for toggling the functionality.

valeriansaliou:
From what I recall from the times I implmented it (my memory might be spotty), it addresses the issue where:

1. You'd have two XMPP clients running at the same time, and you'd actively use one;
[...]

PR #107, opened 2022-08-11
Message carbons

nesium:
This PR adds support for message carbons as mentioned in #46.

That means that if you send or receive a message from or to another client that is logged into the same account, the Prose client will display the message as well but not show a notification for it.


RemiBardon (approved):
LGTM 👍🏼 Thank you 🙏🏼
And thank you for the extensive tests! 🧪

PR #108, opened 2022-08-12
feat: Improve typing indicator

RemiBardon:
For now it looks like:
Screenshot 2022-08-12 at 19 17 44

I will just make it so we don't use "+ n other(s)" is it's not necessary.


valeriansaliou:
LGTM, multi-line is not needed, we'll keep everything single-line for UI reasons.

RemiBardon:
> multi-line is not needed

That's why I have to figure out how to force everything to fit in one line 🥲
I'm almost there, on Monday I won't work so it should be done by Tuesday at noon.

Prose/ProseLib/Sources/ConversationFeature/MessageBar/TypingIndicator.swift
```diff
@@ -0,0 +1,205 @@
...
+//
+
+import AppLocalization
+import Assets
+import ProseCoreTCA
+import SwiftUI
+
+private let l10n = L10n.Content.MessageBar.self
+
+// TODO: Find a better name
```
RemiBardon, line 13: @nesium How could I name this?

RemiBardon: `Name` maybe?

nesium: I assume that we'll just have a `User` there at some point. Say a `Chat` had a dictionary `let participants: [JID: User]` and the `participantStates` contains the `JID`, we'd look up the matching participant and would take the name from there. So, right now I don't think we need to worry about the somewhat weird name.

nesium (approved):
Looks good to me! Thanks! 👍

PR #109, opened 2022-08-12
Load one page of messages from archive

nesium:
Quick & dirty implementation really.

When a conversation appears the last page of messages from that conversation is loaded. Cached messages are discarded, which will make it easier to find any potential bugs so that we know it wasn't cached.


valeriansaliou:
That's exciting!

nesium:
> LGTM 👌🏼
> 
> How soon do you need me to implement [prose-core-views `0.12.0`](https://github.com/prose-im/prose-core-views/releases/tag/0.12.0)? This way we'll be able to load older pages.

Not too soon actually. One reason why this is a quick & dirty implementation is because I'm still uncertain how the mix between Rust and Swift should actually look. I hope to gain some more clarity by implementing a couple of more features. So support for editing messages would be preferable from my side.

valeriansaliou:
> > LGTM 👌🏼
> > How soon do you need me to implement [prose-core-views `0.12.0`](https://github.com/prose-im/prose-core-views/releases/tag/0.12.0)? This way we'll be able to load older pages.
> 
> Not too soon actually. One reason why this is a quick & dirty implementation is because I'm still uncertain how the mix between Rust and Swift should actually look. I hope to gain some more clarity by implementing a couple of more features. So support for editing messages would be preferable from my side.

Let me know if you need any help on the Rust code structure.

PR #110, opened 2022-08-16
feat: Dynamically show typing indicator

RemiBardon:


Prose/ProseLib/Sources/ConversationFeature/MessageBar/MessageBarTextField.swift
```diff
@@ -62,42 +69,76 @@ struct MessageBarTextField: View {
...
   switch action {
   case .sendTapped where !state.message.isEmpty:
     let content = state.message
     state.message = ""
     return Effect(value: .send(message: content))
 
-  default:
+  case .binding(\.$isFocused) where state.isFocused && !state.message.isEmpty,
+       .binding(\.$message) where !state.message.isEmpty:
+    return .merge([
```
RemiBardon, line 93: I always use `.concatenate`, but this time I told myself that we don't need the two effects to run one after another, so I used `.merge`.

Should we go for `.merge` by default unless we need effects to run in a certain order or go for `.concatenate` by default unless we need effects to run in *parallel(?)*?

nesium: The former. Run in parallel unless there's a good reason (a dependency most likely) not to.

Prose/ProseLib/Sources/ConversationFeature/MessageBar/MessageBarTextField.swift
```diff
@@ -62,42 +69,76 @@ struct MessageBarTextField: View {
...
+  ConversationEnvironment
+> = Reducer { state, action, environment in
   switch action {
   case .sendTapped where !state.message.isEmpty:
     let content = state.message
     state.message = ""
     return Effect(value: .send(message: content))
 
-  default:
+  case .binding(\.$isFocused) where state.isFocused && !state.message.isEmpty,
```
nesium, line 91: What do you think about this alternative reducer implementation? …

It's not less code, quite the opposite. However when the reducer acts on a change of state I often find it easier to user `onChange(of:)` instead of switching over actions + state.

```swift
public let messageBarTextFieldReducer: Reducer<
  MessageBarTextFieldState,
  MessageBarTextFieldAction,
  ConversationEnvironment
> = Reducer { state, action, environment in
  switch action {
  case .sendTapped where !state.message.isEmpty:
    let content = state.message
    state.message = ""
    return Effect(value: .send(message: content))

  case let .setChatState(chatState):
    return environment.proseClient
      .sendChatState(state.chatId, chatState)
      .fireAndForget()

  case .onDisappear:
    return .cancel(token: MessageBarTextFieldEffectToken.self)

  case .sendTapped, .send, .binding:
    return .none
  }
}
.binding()
.onChange(of: TextFieldState.init) { current, state, action, environment in
  switch (current.isFocused, current.message.isEmpty) {
    // If the textfield is not focused or empty we don't consider the user typing.
    case (false, _), (_, true):
      return .merge([
        environment.proseClient
          .sendChatState(state.chatId, .active)
          .fireAndForget(),
        .cancel(id: MessageBarTextFieldEffectToken.pauseTyping),
      ])
    
    case (true, _), (_, false):
      return .merge([
        environment.proseClient
          .sendChatState(state.chatId, .composing)
          .fireAndForget(),
        Effect(value: .setChatState(.paused))
          .delay(for: .pauseTypingDebounceDuration, scheduler: environment.mainQueue)
          .eraseToEffect()
          .cancellable(id: MessageBarTextFieldEffectToken.pauseTyping, cancelInFlight: true),
      ])
  }
}

private struct TextFieldState: Equatable {
  var message: String
  var isFocused: Bool
  
  init(_ state: MessageBarTextFieldState) {
    self.message = state.message
    self.isFocused = state.isFocused
  }
}
```

Prose/ProseLib/Sources/ConversationFeature/MessageBar/MessageBar.swift
```diff
@@ -126,22 +158,31 @@ public let messageBarReducer: Reducer<
...
     self.typing = typing
+    self.textField = .init(chatId: chatId, chatName: chatName, message: message)
   }
 }
 
 // MARK: Actions
 
 public enum MessageBarAction: Equatable, BindableAction {
+  case onAppear, onDisappear
+  case typing([Name])
```
nesium, line 185: `.typing` is incredibly terse. If you derive the name from what actually happens that would be `.participantChatStatesChanged` or after mapping and filtering `.composingUsersChanged`.

Prose/ProseLib/Sources/ConversationFeature/MessageBar/MessageBar.swift
```diff
@@ -126,22 +158,31 @@ public let messageBarReducer: Reducer<
...
   var reactionPicker: ReactionPickerState?
 
   public init(
-    textField: MessageBarTextFieldState,
-    typing: [TypingUser] = []
+    chatId: JID,
+    loggedInUserJID: JID,
+    chatName: Name,
+    message: String = "",
+    typing: [Name] = []
```
nesium, line 172: Same here. `typing` could also be a `Boolean`. Maybe `typingUsers` or `composingUsers` if we want to stick to one terminology?

Prose/ProseLib/Sources/ConversationFeature/MessageBar/MessageBarTextField.swift
RemiBardon: This is better, you're right. I'm not used to using `.onChange` but I should definitely think about it more often.

I just changed `switch (current.isFocused, current.message.isEmpty)` for `if !current.isFocused || current.message.isEmpty`.

nesium (approved):
👍

PR #113, opened 2022-08-18
feat: Allow message edition

RemiBardon:
The sheet looks like this:
Screenshot 2022-08-19 at 12 52 24

The placeholder is the initial text:
Screenshot 2022-08-19 at 12 52 41

The "Save" button only shows when changes were made:
Screenshot 2022-08-19 at 12 52 32

It's localized.


valeriansaliou:
Noticed two issues w/ edition:

1. An edited message comes down in the message list, even if it was not the latest message in the history. It should stay in its original position.
2. Probably unrelated, but issue occurs on this branch: the message field does not catch Enter keypresses anymore. When I press enter, it now inserts a new line (which should only be the case if I press SHIFT+ENTER)

RemiBardon:
I wanted to mention the line break, it’s because of TextEditor and there is nothing I can do about it until we introduce the NSTextView wrapper. I thought about conditionally showing a TextField or TextEditor depending if we’re editing the message, but this felt too cumbersome.

For the message placement, I hadn’t seen it. It reminds me of a bug caused by the diffing for ChatView but I thought I had fixed it. I’ll look into it this afternoon.

RemiBardon:
And for the CI I suppose I miss typed a space just before committing…

valeriansaliou:
Mhh on the line break I'm confused, because this works in `master` as of now. Something definitely broke in the `edit-message` branch regarding sending messages w/ the press of one ENTER.

RemiBardon:
> An edited message comes down in the message list, even if it was not the latest message in the history. It should stay in its original position.

I don't see this locally 🤔

RemiBardon:
I fixed the return key, by introducing a boolean to show either a `TextField` (single line, return sends the message) or a `TextEditor` (multi line, no support for return key).

In the message edition sheet, I use a multiline field, and the user has to explicitly hit "Save" to save their edits. I think it makes sense.

Also, I put a `minHeight` on the sheet, and not a `height`, which gives this gorgeous feature:

https://user-images.githubusercontent.com/37386490/185644111-430f57a9-5e41-47e8-8521-d44eb1dd4f1f.mov

Of course, the size resets when the sheet is opened again, not giving a weird size when not needed:

https://user-images.githubusercontent.com/37386490/185644546-d212a483-1992-4588-8f5b-bc0be22d4875.mov

Notice the great animation too:

https://user-images.githubusercontent.com/37386490/185644761-ba84f42a-3274-4a87-9259-eb3a33b28fdb.mov

Prose/ProseLib/Tests/ProseCoreTCATests/ProseClientTests.swift
```diff
@@ -403,7 +403,7 @@ final class ProseClientTests: XCTestCase {
     let chatMessages = TestSink<IdentifiedArrayOf<Message>>()
     var c = Set<AnyCancellable>()
 
-    try self.await(client.sendMessage("chat1@prose.org", "message 1"))
+    try self.await(client.sendMessage("chat1@prose.org", "message 0"))
```
RemiBardon, line 406: @nesium I only changed the messages/carbons contents, to help understanding, but I didn't change the messages/carbons IDs to not alter the test case.

Prose/ProseLib/Sources/ProseCore/ProseClient.swift
```diff
@@ -65,11 +65,11 @@ public final class ProseClient: ProseClientProtocol {
 
   public func updateMessage(
     id: String,
-    newID: String,
+    carbonId: String,
```
nesium, line 68: Please revert all instances where you've changed `newID` to `carbonId`. When you update a message you're not creating a message carbon but a new message. Only the server creates carbons. They are semantically different things.

nesium (commented):
Looks good! Other than the small change I've mentioned in the comment.

Prose/ProseLib/Sources/ProseCore/ProseClient.swift
RemiBardon: I don't agree with you because technically we're not updating the message ID, which is implied by `updateMessage`/`newID`. It was the reason for my initial change but I reverted because it's not important.

nesium: I agree that the method signature is confusing. It's a leaky abstraction over XMPP details. `carbonId` would still be flat out wrong.

nesium (approved):

Issue #129, opened 2023-05-30
Login failure with valid credentials


RemiBardon:
When I try to log into Prose (`main@490db37`), with valid credentials, I get:

```log
request error: <error type="cancel"><item-not-found xmlns="urn:ietf:params:xml:ns:xmpp-stanzas"/></error>
```

@valeriansaliou Do you have an idea what's going on? Is my account in a bad state as I haven't logged in for a while?

RemiBardon:
I am logged into Prose web at the same time, maybe it's important to note.

nesium:
If you activate the `PROSE_CORE_LOG_ENABLED` env var and set the `PROSE_CORE_LOG_LEVEL` env var to `TRACE` you should be able to see the data exchange with the server in the Console.app. Could you share the last few messages before the failure occurs? It also happen when the vCard is loaded. That is meant to fail gracefully, but maybe there's a bug.

RemiBardon:
Yes! It didn't cross my mind
I will try soon and send you the logs 🪵

nesium:
Hey Rémi. Could you try again with the latest changes on master? The core lib now includes your patch and I've fixed a couple of issues that affected the first login.

RemiBardon:
On it

RemiBardon:
Seems fixed, I was able to log in and "introduce myself"

PR #130, opened 2023-05-30
chore: Fix warning (non-optional value compared to `nil`)

RemiBardon:
While building the app, I got:

```log
Comparing non-optional value of type 'ConversationInfoReducer.State' (aka 'ChatSessionState<ConversationInfoReducer.ConversationInfoState>') to 'nil' always returns false
```

Now that `info` is initialized when the state is created, I don't think this code is necessary at all anymore. It may be useful when switching between conversations, we must try this before merging (hence the Draft status). I couldn't try locally because of #129, so can one of you confirm it doesn't break anything?


nesium (approved):
Thanks for adding the test, Rémi!

RemiBardon:
You're welcome! I need to get used to it 😌

